PR #1, opened 2019-11-14
Remove postdeploy:postdeploy

prrrnd:
This task is not needed as the Procfile declares a release command.
Heroku's new review apps changes also make it impossible to predict
a review app's name, which makes this task obsolete.


GeoffAbtion:
> Heroku's new review apps changes also make it impossible to predict
a review app's name

It must be possible for the postdeploy hook to get the review app name somehow?
(theoretically. I've not read the docs....)

PR #2, opened 2019-11-14
Fix undefined method `start_with?' for /\.(?:svg|eot|otf|woff|ttf)$/

prrrnd:
Sprockets no longer support regex as a precompile option.

Ref: https://github.com/rails/sprockets/issues/632


hcarreras:
👏

PR #3, opened 2019-11-14
Add a basic service objects set up

prrrnd:
Add app/services with a base class and an example.
Add a result object, that all services should return.

### DIFF
--- a/spec/services/example_service_spec.rb
+++ b/spec/services/example_service_spec.rb
@@ -1,0 +1,19 @@
+require "rails_helper"
+
+RSpec.describe ExampleService, type: :service do
+  describe "#execute" do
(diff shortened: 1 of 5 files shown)


spec/services/example_service_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "rails_helper"
```
jeppester: I looks like this is missing the obnoxious `# frozen_string_literal: true`-comment

prrrnd: Good catch. I just realized it's actually not enforced. Will open a new PR for it :-)

hcarreras:
👏

PR #4, opened 2019-11-14
Register headless chrome as a new Capybara driver

prrrnd:
Headless chrome is now the default. To use the previous driver, set:
TEST_DRIVER=chrome in .env.test

This PR does not cover extracting the capybara configuration into its own file, as mentioned on the Asana board by @GeoffAbtion.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -38,8 +38,8 @@ Capybara.register_driver :headless_chrome do |app|
   Capybara::Selenium::Driver.new(app, browser: :chrome, options: browser_options)
 end
 
-Capybara.javascript_driver = ENV.fetch("TEST_DRIVER", "headless_chrome").to_sym
-Capybara.default_driver = ENV.fetch("TEST_DRIVER", "headless_chrome").to_sym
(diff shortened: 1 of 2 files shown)


spec/spec_helper.rb
```diff
@@ -29,8 +29,17 @@
...
+  Capybara::Selenium::Driver.load_selenium
+  browser_options = ::Selenium::WebDriver::Chrome::Options.new
+  browser_options.args << "--headless"
+  browser_options.args << "--disable-gpu"
+  browser_options.args << "--window-size=1920,1080"
+  Capybara::Selenium::Driver.new(app, browser: :chrome, options: browser_options)
+end
+
+Capybara.javascript_driver = ENV.fetch("TEST_DRIVER", "headless_chrome").to_sym
+Capybara.default_driver = ENV.fetch("TEST_DRIVER", "headless_chrome").to_sym
```
jeppester: Not a big thing, but I think `CAPYBARA_DRIVER` would be a more obvious name for the env var

hcarreras:
👏

PR #5, opened 2019-11-14
Update rubocop's Style/FrozenStringLiteralComment to 'always'

prrrnd:


runephilosof-abtion:
@GeoffAbtion The commits were squashed in the end...
I agree, that often it is nicer to have multiple commit.
However, it also comes with the disadvantage that going back to the first commit the tests aren't green (if we include linting tests in the definition).
Also, if you need to revert, then one commit is nicer.

GeoffAbtion:
When reviewing a PR seeing small logical atomic commits one at a time is nice.
So it is good to separate manual changes (the config file edit.) with automatic changes (the `rubocop -a`)
The commits tell the story of how the end state was reached.
The result of the PR is all that matters for test passing.

GeoffAbtion:
When you are doing "true" TDD you will not have a situation where a commit is red, and you can still tell a story with your commit messages.
When you are messing about with linter configurations and need to separate the manual change from automatic fixes you will 🤷‍♀ 
You can never have too many commits, most people have too few.

PR #6, opened 2019-11-19
Support DATABASE_URL in .env.local.

runephilosof-abtion:
It doesn't set the database name from `DATABASE_URL`, so you can use the same url for both development and test.


prrrnd:
Thanks for bringing this back into this project 👍

PR #7, opened 2019-11-26
Add basic structure for a JSON API with user authentication

prrrnd:
This commit sets a basic structure for a JSON API and handles:
* API versioning via a HTTP header
* User authentication via a HTTP header
* JSON serialization via fast_jsonapi

Shortcomings:
* To keep it as simple as possible, I opted for a single authentication token per user
* The token is not set to expire

I would like some feedback on this proposal, and would write documentation on API versioning and authentication if accepted.
I think this level is the right balance between being easily understandable by more junior developers, and setting a good foundation for more advanced developers.

### DIFF
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -8,7 +8,9 @@ module Api
       private
 
       def require_authentication!
-        user = User.find_by(authentication_token: request.headers["X-Api-Key"])
+        user = authenticate_or_request_with_http_token do |token, _options|
+          User.find_by(authentication_token: token)
+        end
 
         if user.present?
           sign_in user
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class BaseController < ApplicationController
+      skip_before_action :verify_authenticity_token
+
+      private
+
+      def require_authentication!
+        user = User.find_by(authentication_token: request.headers["X-Api-Key"])
+
+        if user.present?
+          sign_in user
+        else
+          render status: :unauthorized
+        end
+      end
+    end
+  end
+end
--- a/app/models/concerns/api_authenticable.rb
+++ b/app/models/concerns/api_authenticable.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module ApiAuthenticable
+  AUTHENTICATION_TOKEN_LENGTH = 64
+
+  extend ActiveSupport::Concern
+
+  included do
+    before_create do
+      self.authentication_token = generate_authentication_token
+    end
+
+    def generate_authentication_token
+      loop do
+        token = Devise.friendly_token(AUTHENTICATION_TOKEN_LENGTH)
+        break token unless self.class.find_by(authentication_token: token)
+      end
+    end
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -6,7 +6,7 @@ Rails.application.routes.draw do
   devise_for :users
 
   namespace :api, defaults: { format: :json } do
-    scope module: :v1, constraints: Constraints::ApiVersion.new("v1", true) do
+    scope module: :v1, constraints: Constraints::ApiVersion.new("v1") do
       resources :users, only: :show
     end
   end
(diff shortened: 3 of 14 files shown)


app/controllers/api/v1/base_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module Api
+  module V1
+    class BaseController < ApplicationController
+      skip_before_action :verify_authenticity_token
+
+      private
+
+      def require_authentication!
+        user = User.find_by(authentication_token: request.headers["X-Api-Key"])
```
runephilosof-abtion: I suggest using `Authorization: Bearer <token>`. It will also make some log services filter it out automatically.

app/models/concerns/api_authenticable.rb
```diff
@@ -0,0 +1,20 @@
...
+module ApiAuthenticable
+  AUTHENTICATION_TOKEN_LENGTH = 64
+
+  extend ActiveSupport::Concern
+
+  included do
+    before_create do
+      self.authentication_token = generate_authentication_token
+    end
+
```
runephilosof-abtion: Do we always want to create an authentication_token for the users?
How about just supplying the method to create it, but let it up to the controllers to actually create it when needed?

hcarreras: I believe then it would have a high risk of being overlooked. 
It's fairly easy to remove it if it's not being done. I believe we are going to have this discussion repeated times in the future due to the structure of this template.

spec/serializers/user_serializer_spec.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe UserSerializer, type: :serializer do
+  it "only exposes a subset of the user's data" do
+    user = FactoryBot.create(:user)
```
hcarreras: Do we need to specify `FactoryBot`?

jeppester:
The code here looks fine to me, I feel though like there's missing some functionality allowing a user to get the token.

GeoffAbtion:
> Even though I like that we are setting fast_json_api as a default, it's worth noting that we haven't used it (that I know of) in any project.

We're using it on Selfcast

GeoffAbtion:
This PR does too many things, which makes it hard to review :(

config/routes.rb
```diff
@@ -1,7 +1,15 @@
 # frozen_string_literal: true
 
+require "constraints/api_version"
+
 Rails.application.routes.draw do
   devise_for :users
 
+  namespace :api, defaults: { format: :json } do
+    scope module: :v1, constraints: Constraints::ApiVersion.new("v1", true) do
```
GeoffAbtion: Boolean method argument are obscure, how do I know what `true` means here?

Show the meaning of the argument in the thing you are calling something like:

`Constraints::ApiVersion::Default.new("v1")`

or

`Constraints::ApiVersion.default("v1")`

lib/constraints/api_version.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Constraints
+  class ApiVersion
+    attr_reader :version, :default
+
+    def initialize(version, use_default = false)
```
GeoffAbtion: Removing the boolean argument also greatly simplifies everything in here.

GeoffAbtion:
I really don't like adding the token as a property of the `User` model.

I can see the simplicity trade off here, but I think Rails gives us the tools to do this in a nicer way.

Our theoretical junior developers are going resist splitting token authentications out into their own table and will want to add more stuff to the `User` table (don't do this.)

But maybe this is all YAGNI anyway 🤷‍♂

config/routes.rb
GeoffAbtion: or `Constraints::DefaultAPiVersion("v1")`

app/models/concerns/api_authenticable.rb
prrrnd: I agree with Hari here, this is easy to turn off. I would assume it is needed in most cases.

heatherm:
@prrrnd Can you respond to my thoughts on using the token auth gems? Curious why we need to build our own wheel here? Also curious about defaulting the template to token auth-- I think this makes a bit of an assumption about the types of projects (see: mobile front end), that I'm not sure I see supported from our sales efforts.

PR #8, opened 2019-11-28
Remove conventions section

jeppester:
I don't think every project following the template needs a section about how to make code blocks look better in markdown.
If adding a convention about this is necessary, I think it belongs somewhere else, like for instance in the employee manual.


prrrnd:
Agreed, thanks for the contribution 👍  I'll leave this open for others to chip in, if there's any need to discuss this.

hcarreras:
I know there was some discussion about the information that the README should have in case of a handover. 
I don't hold a strong opinion. So, I'll stay neutral. If you still think it's good to remove the section taking into account that the project might be handed over to another company, I'm fine with this.

PR #9, opened 2019-11-28
Remove Heroku section

jeppester:
I don't think having a section like this in the project adds any value on a day-to-day basis.
It does however add the burden of keeping the prices updated, and the actual heroku location, db plans etc. in sync with what's in the README.

I understand if the point of the section was to guide developers on how to set up heroku when initially creating the project. In that case let's add such a section to the template's README instead.

### DIFF
--- a/README.example.md
+++ b/README.example.md
@@ -164,11 +164,13 @@ Exclusions:
 
 # Production
 
-## Deployments
-
 The project is hosted by [heroku](https://heroku.com).
 
-Review apps and CI are enabled for PR's. Auto merge setup to staging environment on merging of branches into master.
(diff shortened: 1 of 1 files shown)


jeppester:
@heatherm I guess I thought it would be obvious where to find more information, given that it's mentioned that the project is on heroku, and the project name is part of the git repo (not thinking too much about PMs I guess :smile: ).

Would it be enough to add a link to `https://dashboard.heroku.com/apps/project-name` and mention that access must be granted for the abtion team?

I don't have any issues with pointing people towards heroku, what I don't want is having to keep redundant information up-to-date, and I'm 100% sure that we won't be able to live up to that.

heatherm:
@jeppester sounds good.

jeppester:
I've updated the PR, please review again

PR #13, opened 2019-12-04
Reference to Guidelines README as a start

hcarreras:
Currently, our README is duplicated in both the Muffi template and the guidelines


hcarreras:
I've realized that the READMEs don't much exactly. I'll think of a solution before reopening this PR. Suggestions re welcome

prrrnd:
I'd like to see a "templates" repository, and then a command to add templates to new projects, perhaps as simple as 
```sh
curl -fsSL https://raw.githubusercontent.com/abtion/templates/master/README.md > README.md
```

This would also allow us to have different versions with branches / tags.

hcarreras:
It sounds like a good idea, but It also adds the complexity of keeping up to date several branches. So, anytime we do a change in a README, we have to remember to apply the changes in all the other branches.

bohme:
Int he spirit of #15 having a generator take care of the README strikes me as a good solution.
You can have the template as an erb file and use conditionals to choose what gets included. 🤔

heatherm:
`README.md.tt` in the template gets my vote.

README.md
```diff
@@ -47,7 +47,7 @@ your own repository.
 4. `cp .env.example .env`
 5. `cp .env.development.example .env.development`
 6. `cp .env.test.example .env.test`
-7. `mv README.example.md README.md`
+7. `curl -o README.md https://raw.githubusercontent.com/abtion/muffi/master/README.example.md`
```
jeppester: I feel like I'm missing out on something here. If this gets merged, won't the curl command then return a 404?

bohme: Jesper is right :O
is the URL pointing to the wrong repo?

jeppester:
What is the status here? Are we going to update the README in the guidelines? Are we going to fix the link here, and merge?

I don't mind spending some time on this, I just need to know what to do :-)

hcarreras:
Unfortunately, fixing the link won't be enough. 
The issue is that we are trying to have a "universal" template, but a Rails README will be different from a NodeJs Readme. The structure should be similar, but the content will differ.
I see no way of having a single README, so, it's up for discussion what will make it easier to maintain the different READMEs. 
I believe it's a bit better to keep several documents in the Guidelines. 
`README.md.tt`,  `README_RAILS.md.tt` and so on.

heatherm:
I think we can have separate repos for rails and node readmes. Maybe stick to just rails for now to simplify?

heatherm:
I think we should maintain one repo per language template, and that those templates should contain a readme that is a good starting point for a new project in that language.

Thus, muffi here (which is rails) would have a README.md.tt and muffi-node would have it's own that is specific to node. We can take that on when we create a node template someday.

From guidelines, we can link to the individual templates to showcase them and also point to any notable differences between the language-based-readme-needs if that is relevant. For now, this means we would just point to the readme in this rails template, and people starting node projects will need to use their best judgement on what to shift.

This strategy leaves some chance that readmes in the node world don't benefit from readmes in the rails world and vice versa, but I think we should attend to that when it becomes an issue. For now, it's fairly easy for us to just use the rails readme as the base and example (I think).

Thoughts?

jeppester:
@heatherm I agree. Another argument that favors keeping the README in this repo is that the README is absolutely tied to the template - It would be quite odd to use the README without the template or vice versa.

hcarreras:
Cool. Then, I'll close this PR. I'll add examples to the "general" README.

PR #18, opened 2019-12-17
Update minimum password length to 8

hcarreras:
New Abtion security policy.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -3,6 +3,23 @@
 require "rails_helper"
 
 RSpec.describe User, type: :model do
+  describe "password" do
+    context "password 8 characters long" do
+      it "user valid" do
+        user = build(:user, password: "12345678")
+        expect(user).to be_valid
+      end
+    end
+
+    context "password less than 8 characters long" do
+      it "user not valid" do
+        user = build(:user, password: "1234")
+        expect(user).to_not be_valid
+        expect(user.errors[:password]).to eq(["is too short (minimum is 8 characters)"])
+      end
+    end
+  end
+
   describe "admin role" do
     let(:user) { create(:user) }
 
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -11,9 +11,9 @@ RSpec.describe User, type: :model do
       end
     end
 
-        user = build(:user, password: "1234")
     context "without 8 character long password" do
       it "user is not valid" do
+        user = build(:user, password: "1234567")
         expect(user).to_not be_valid
         expect(user.errors[:password]).to eq(["is too short (minimum is 8 characters)"])
       end
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -4,16 +4,16 @@ require "rails_helper"
 
 RSpec.describe User, type: :model do
   describe "password" do
-    context "password 8 characters long" do
-      it "user valid" do
+    context "with 8 character long password" do
+      it "user is valid" do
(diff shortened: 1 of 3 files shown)


heatherm:
> additional password sanity checks (like [pwned password k-anonymity check](https://www.troyhunt.com/ive-just-launched-pwned-passwords-version-2/#cloudflareprivacyandkanonymity))

Added here https://app.asana.com/0/1142794766483633/1154313617630257/f

GeoffAbtion:
> I think we should add a spec validating the error message with a password too short.

Unless we're customising the error message, this would be a test of Rails functionality (which should already be tested by Rails.)

I agree we need a specific test to cover the password length criteria.

(ignoring my objections to shoulda matchers for now)
Something like:
```
describe User do
  it { should validate_length_of(:password).is_at_least(8)
end
```

Whether we also need to test the upper bound is another question.

spec/models/user_spec.rb
```diff
@@ -3,6 +3,23 @@
...
+    context "password 8 characters long" do
+      it "user valid" do
+        user = build(:user, password: "12345678")
+        expect(user).to be_valid
+      end
+    end
+
+    context "password less than 8 characters long" do
+      it "user not valid" do
+        user = build(:user, password: "1234")
```
GeoffAbtion: This test would still pass even without the devise config change.

GeoffAbtion: Always have a failing test before making a change.

spec/models/user_spec.rb
```diff
@@ -3,6 +3,23 @@
 require "rails_helper"
 
 RSpec.describe User, type: :model do
+  describe "password" do
+    context "password 8 characters long" do
```
prrrnd: `context` descriptions should start with `with/out` or `when` to make it more readable.
`it` descriptions should contain a verb.

Something like is easier to read:
```ruby
context "when the password is at least 8 chars long" do
  it "passes the validations" do
    # ...
  end
end
```

http://www.betterspecs.org/#contexts

spec/models/user_spec.rb
prrrnd: Agree, but why would this pass without the config change? It does mention 8 chars in the expectation.

runephilosof-abtion: Before we had 6 characters, so in order to fail before the change, the spec would need to be 6 or 7 characters long.

runephilosof-abtion:
I think we can merge this now that Heathers change request has been fulfilled.

spec/models/user_spec.rb
jeppester: First: I agree that this leads to an improvement when reading the whole concatenated string of describe -> context -> it.

Having said that, I'm not sure that we should use betterspecs as a source of truth when it comes to questions regarding specs. Some of the things it encourages are a bit controversial to me (and not only me I think). For instance:
- Using subject to "dry tests up"
- Using `let`
- Using shared examples

I've seen all of those spiraling completely out of control, and it makes me think if we should have our own appendix to those guidelines, or if there's another guide that we can reference instead - or maybe I'm completely alone about this?

What are your thoughts? @heatherm @prrrnd @hcarreras @GeoffAbtion @runephilosof-abtion

bohme: DRY shouldn't apply to specs in my junior opinion.
Making test be able to stand on their own and be readable is the way to go 👍

GeoffAbtion: I have strong opinions on many of these topics.
Remind me to post something about them when I'm not sick 🤒

PR #19, opened 2019-12-18
Make project-specific strings easier to search-replace

jeppester:
Makes it easier to use search-replace to update project-name strings across all files:
- PROJECT_DOMAIN => project-name.com
- PROJECT_NAME_PARAM => project-name
- PROJECT_NAME_PASCAL => ProjectName
- PROJECT_NAME_SNAKE => project_name
- PROJECT_NAME_HUMAN => Project name

Asana: https://app.asana.com/0/1142794766483633/1152060241410857

NB: I'm not sure how rails templates work, but I guess they have a built-in way to do something like this - in which case this might have to be migrated at some point

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,12 @@ your own repository.
 
 1. Edit `application.rb` and change the module name and configuration settings.
 2. Edit the database names in `database.yml`
-3. Search the project for the word `PROJECT_NAME` to find other things to change.
+3. Use "search and replace" to replace project specific strings across all files:
+   - PROJECT_DOMAIN => project-name.com
+   - PROJECT_NAME_PARAM => project-name
+   - PROJECT_NAME_PASCAL => ProjectName
+   - PROJECT_NAME_SNAKE => project_name
+   - PROJECT_NAME_HUMAN => Project name
 4. `cp .env.example .env`
 5. `cp .env.development.example .env.development`
 6. `cp .env.test.example .env.test`
--- a/README.example.md
+++ b/README.example.md
@@ -167,15 +167,15 @@ By default, only admins are authorized to call a controller action.
 
 The project is hosted by [heroku](https://heroku.com).
 
-Current dyno types and add-on plans can be found in the project's [heroku dashboard](https://dashboard.heroku.com/apps/PROJECT_NAME-production). To access the dashboard, a heroku user with access to the abtion team is required.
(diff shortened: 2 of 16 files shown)


README.md
```diff
@@ -22,12 +22,12 @@ Ruby on Rails applications.
...
-* Uses [Heroku CI](https://heroku.com)
-* Uses [rspec](https://github.com/rspec/rspec-rails)
-* Uses [simplecov](https://github.com/colszowka/simplecov)
-* Uses [rubocop](https://github.com/bbatsov/rubocop)
+- Heroku ready - Review apps, staging and production
+- Uses [devise](https://github.com/plataformatec/devise)
+- Uses [Heroku CI](https://heroku.com)
+- Uses [rspec](https://github.com/rspec/rspec-rails)
+- Uses [simplecov](https://github.com/colszowka/simplecov)
+- Uses [rubocop](https://github.com/bbatsov/rubocop)
```
bohme: unrelated changes?

README.md
```diff
@@ -43,7 +43,12 @@ your own repository.
 
 1. Edit `application.rb` and change the module name and configuration settings.
 2. Edit the database names in `database.yml`
-3. Search the project for the word `PROJECT_NAME` to find other things to change.
+3. Use "search and replace" to replace project specific strings across all files:
+   - PROJECT_DOMAIN => project-name.com
+   - PROJECT_NAME_PARAM => project-name
+   - PROJECT_NAME_PASCAL => ProjectName
+   - PROJECT_NAME_SNAKE => project_name
+   - PROJECT_NAME_HUMAN => Project name
```
bohme: I'd have thought these would just be correctly cased versions of `PROJECT_NAME`.
As in `ProjectName`, `project-name`, `project_name` and `Project name` 🤔 
then one could use smart search/replace to convert all of them at once.
Any reason you did it this way?

README.md
jeppester: I tried my best to not include vs code's automatic formatting... It clearly wasn't enough!

README.md
jeppester: That was my initial idea, but then when I got to the domain, I got worried that two differently used strings could look the same but not necessary have to be the same.

That problem might be hard to spot now because I later decided to include the `.com`-part in the domain, but I think it's still a potential issue.

For instance at some point we might want to have a "PROJECT_NAME_HEROKU", it might often be "project-name" but it's not reliably the same.

PR #21, opened 2019-12-19
README: Merge sections `developing` and `development`

jeppester:
This is an attempt to simplify the README and remove redundant and sometimes ambiguous information (should the server run with `heroku local` or `bundle exec rails s`?)

All the information in this PR is from either the previous `developing` section or the previous `development` section, and I'm just trying to tie them together. So I hope I'm not stepping on someones toes. If that's the case however I'm completely open for suggestions.

### DIFF
--- a/README.example.md
+++ b/README.example.md
@@ -2,46 +2,27 @@
 
 This project is built on top of [Muffi](https://github.com/abtion/muffi).
 
-## Developing
-
-### Getting Started
-
-* After cloning this repo, run: `bin/setup`
-
-### Day-to-day
-
-* Run the server: `heroku local` and [http://localhost:5000](http://localhost:5000)
-* Run tests: `rspec`
-
-### Debugging
-
-* Call `byebug` anywhere in the code to stop execution and get a debugger console.
-* Access an IRB console on exception pages or by using `<%= console %>` anywhere in the code.
-* (Of course, [RubyMine](https://www.jetbrains.com/ruby/) includes a great [visual debugger](https://www.jetbrains.com/ruby/features/ruby_debugger.html)).
-
-### HerokuCI green builds deploy to Heroku staging servers
-
 # Name of the project
 
 1. [README](#readme)
-   1. [Developing](#developing)
-      1. [Getting Started](#getting-started)
-      2. [Day-to-day](#day-to-day)
-      3. [Debugging](#debugging)
-      4. [HerokuCI green builds deploy to Heroku staging servers](#herokuci-green-builds-deploy-to-heroku-staging-servers)
 2. [Name of the project](#name-of-the-project)
-3. [Development](#development)
-   1. [Requirements](#requirements)
-   2. [Set up the app the first time](#set-up-the-app-the-first-time)
-   3. [Run the app](#run-the-app)
-   4. [Run the tests](#run-the-tests)
-   5. [Download production or staging DB](#download-production-or-staging-db)
-   6. [Notable inclusions and Notable exclusions](#notable-inclusions-and-notable-exclusions)
-      1. [Devise User Authorization](#devise-user-authorization)
-4. [Production](#production)
+3. [Requirements](#requirements)
+4. [Developing](#developing)
+   1. [First time setup](#first-time-setup)
+      1. [1. Configuration](#1-configuration)
+         1. [Database connection](#database-connection)
+      2. [2. Dependencies and database setup](#2-dependencies-and-database-setup)
+         1. [Chrome driver](#chrome-driver)
+      3. [3. Ensure that tests pass](#3-ensure-that-tests-pass)
+   2. [Day-to-day](#day-to-day)
+   3. [Debugging](#debugging)
+   4. [Download production or staging DB](#download-production-or-staging-db)
+5. [Notable inclusions and Notable exclusions](#notable-inclusions-and-notable-exclusions)
+   1. [Devise User Authorization](#devise-user-authorization)
+6. [Production](#production)
    1. [Deployments](#deployments)
-5. [Staging](#staging)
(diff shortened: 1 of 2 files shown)


README.example.md
```diff
@@ -58,8 +36,19 @@ This section should include any business related explanation that helps understa
 
 Name, and a short description of any services that the project is using (error tracking, activity monitoring, log registry, email service, etc). Include where to get the credentials. E.g., 1Password under admin+project@abtion.com.
 
+# Requirements
+
+You must have the following installed and available on your machine:
+
+- **Ruby 2.6.5**
+- **Node JS 12.x**
+- **PostgreSQL 11.3**
```
runephilosof-abtion: I prefer to check the version of the Heroku add-on instance that is connected to the production app.
To be sure you get the right version.

README.example.md
```diff
@@ -58,8 +36,19 @@ This section should include any business related explanation that helps understa
 
 Name, and a short description of any services that the project is using (error tracking, activity monitoring, log registry, email service, etc). Include where to get the credentials. E.g., 1Password under admin+project@abtion.com.
 
+# Requirements
+
+You must have the following installed and available on your machine:
+
+- **Ruby 2.6.5**
+- **Node JS 12.x**
```
runephilosof-abtion: Add yarn?

README.example.md
```diff
@@ -77,69 +66,30 @@ stub any external calls using the test suite's features.
...
-
-```sh
-$ bundle install
-$ npm install -g yarn
-$ yarn install
-```
-
-Prepare the database
-
-You can set `DATABASE_URL` in `.env.local`, if for instance you use Docker for Postgres. `DATABASE_URL="postgresql://localhost:5432"`
```
runephilosof-abtion: Please keep this.

README.example.md
```diff
@@ -77,69 +66,30 @@ stub any external calls using the test suite's features.
...
-```sh
-$ bundle exec rails db:setup
-```
-
-Download the correct version of chromedriver, it will try to do it automatically, but if you disable network with webmock/vcr your tests will fail when it does.
-```sh
-$ bundle exec rails webdrivers:chromedriver:update
-```
```
runephilosof-abtion: I added this, because the error you get isn't intuitive, so you might waste time trying to figure out what to do about it.

README.example.md
```diff
@@ -77,69 +66,30 @@ stub any external calls using the test suite's features.
...
-
-Run the test suite and expect all tests to pass
-
-```sh
-$ bundle exec rspec
-```
```
runephilosof-abtion: You probably deleted this because it is also under Day-to-day.
But it should also be done during first time setup.

README.example.md
```diff
@@ -77,69 +66,30 @@ stub any external calls using the test suite's features.
...
-```sh
-heroku local
-```
+Run: `bin/setup`
 
-## Run the app
+## Day-to-day
 
-`bundle exec rails s`
+- Run the server: `heroku local` and [http://localhost:5000](http://localhost:5000)
```
bohme: are we actually supposed to run it with `heroku local`? 🤔

bin/setup
```diff
@@ -13,12 +13,13 @@ FileUtils.chdir APP_ROOT do
   # This script is idempotent, so that you can run it at anytime and get an expectable outcome.
   # Add necessary setup steps to this file.
 
-  puts '== Installing dependencies =='
+  puts '== Installing ruby dependencies =='
```
bohme: nice 👍

README.example.md
runephilosof-abtion: I have never used it.
How many of you are using `heroku local`, `foreman` or the like?
How well does it combine with using `byebug`?

README.example.md
jeppester: Good catch :+1: , I didn't mean to remove this.

README.example.md
jeppester: Good point! But lets address it in another PR, I want this PR to be mostly about merging the two similar sections in a meaningful way. I'm trying my best to resist the temptation to rewrite a whole bunch of stuff.

README.example.md
jeppester: Good point, it makes sense to add yarn to the dependencies now that we are removing the instructions about how to install it :+1:

README.example.md
jeppester: I decided to go with `heroku local` because it was mentioned before `bundle exec rails s`, and because I think that the person who decided to add `heroku local` must have had an opinion about it.

Personally I don't care a lot.

README.example.md
jeppester: Good point :+1:

README.example.md
jeppester: I removed this because under "configuration files" we also have:
> There are two main configuration files that are of concern:
> 
> - `.env` - configuration that's common across all environments that you intend
>  to run on your machine (e.g. database connection)

In hindsight I agree that it perhaps wasn't the right decision. I think I might move this to "Database setup" and reference it from the part about the configuration file.

jeppester: I've added this again, under "Configuration"

jeppester:
@runephilosof-abtion I believe I've addressed the "blocking" issues now

bin/setup
```diff
@@ -13,12 +13,13 @@ FileUtils.chdir APP_ROOT do
...
   system! 'gem install bundler --conservative'
   system('bundle check') || system!('bundle install')
 
   # Install JavaScript dependencies
```
hcarreras: The comment is no longer necessary as we are saying the same in the next line. I'll remove it :)

hcarreras:
Awesome!

PR #23, opened 2020-01-07
Update placeholder to required syntax

DeWinther:
A name with spaces produces `error package.json: Name contains illegal characters`

Co-authored-by: Heather Moore <hm@abtion.com>


jeppester:
:+1:

PR #24, opened 2020-01-07
Add suffix to .env.test file and gitignore it

DeWinther:
This brings it in line with the readme instructions and correct naming.


prrrnd:
Shouldn't the `.env.test` be in the repo for the CI?

DeWinther:
I'm not entirely sure if that is needed, I just don't think we should push files that are different for all our dev environments?

prrrnd:
I rarely use anything else than just a `.env` file, so I'm not sure how others use them. For the development environment, I would assume we use `.env.development`, but the `.env.test` should be the same for everyone.

DeWinther:
I normally also only use the `.env` file, i just base my comment on the step 8 in Configure the project section in the readme:
> Add export DATABASE_URL="postgresql://HOST:@localhost:5432" to .env.development and .env.test

runephilosof-abtion:
https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use
We should have generally usefull .env and .env.local and use `.env.local`, `.env.test.local` for local changes.

prrrnd:
As Rune mentioned, we should use .local for local changes. I'm going to close this one, but feel free to re-open if you're not happy with the decision :muscle:

DeWinther:
@prrrnd This sounds like a good idea, we then just have to fix the readme so it doesn't suggest adding personal/individual configurations to a file we are uploading

PR #28, opened 2020-01-08
Fix Code Climate environment variable name

runephilosof-abtion:
https://app.asana.com/0/1142794766483633/1156152748855916

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ your own repository.
 2. Go to `https://codeclimate.com/repos/<your-project>/settings/test_reporter`
 3. Set up a Heroku pipeline.  
 4. `heroku config:set CC_TEST_REPORTER_ID=<Code Climate test reporter ID>` 
-4. Update https://codeclimate.com/repos/REPO_ID/settings/test_reporter to be at 100% diff coverage
+5. Update https://codeclimate.com/repos/REPO_ID/settings/test_reporter to be at 100% diff coverage
 
 ### Configure CD
 
--- a/README.md
+++ b/README.md
@@ -59,7 +59,8 @@ your own repository.
 
 1. Login to CodeClimate and add your project
 2. Go to `https://codeclimate.com/repos/<your-project>/settings/test_reporter`
-3. Set up a Heroku pipeline test run config var `CC_TEST_REPORTER_ID` with the Code Climate test reporter ID
+3. Set up a Heroku pipeline.  
+4. `heroku config:set CC_TEST_REPORTER_ID=<Code Climate test reporter ID>` 
 4. Update https://codeclimate.com/repos/REPO_ID/settings/test_reporter to be at 100% diff coverage
 
 ### Configure CD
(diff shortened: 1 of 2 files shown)


runephilosof-abtion:
We could also instead instruct to set up the variable in the Heroku settings for the pipeline.
Is the token a secret?
If it is, then we should definitely not put it in `app.json`

prrrnd:
The name is actually used by Code Climate's script `cc-test-reporter` and is important. We cannot just rename it.

It is supposed to be secret though, so we should find a solution for not having it in the app.json.

hcarreras:
From Heroku documentation: 
> To set confidential or volatile environment variables (such as access tokens) that you shouldn’t include in app.json, you can instead add them to your pipeline’s Heroku CI settings in the Heroku Dashboard:

runephilosof-abtion:
> The name is actually used by Code Climate's script cc-test-reporter and is important. We cannot just rename it.
Actually, it is the opposite. The new variable name is used by cc-test-reporter (https://docs.codeclimate.com/docs/configuring-test-coverage), the old name was used in a ci script in the project that we copied this from.

This is the reason why we need to set up Code Climate, CI and a Heroku app for Muffi. To make sure these things are set up correctly. And to demonstrate to new projects, how it should be set up.

runephilosof-abtion:
> It is supposed to be secret though, so we should find a solution for not having it in the app.json.

I have changed `app.json` to reflect that.

README.md
```diff
@@ -59,7 +59,7 @@ your own repository.
 
 1. Login to CodeClimate and add your project
 2. Go to `https://codeclimate.com/repos/<your-project>/settings/test_reporter`
-3. Copy the test reporter ID to `CODE_CLIMATE_TEST_REPORTER_TOKEN` in `app.json`
+3. Set up a Heroku pipeline test run config var `CC_TEST_REPORTER_ID` with the Code Climate test reporter ID
```
hcarreras: I'm not sure if I understand this line.
```suggestion
3. Set up a Heroku pipeline.  
4. `heroku config:set CC_TEST_REPORTER_ID=<Code Climate test reporter ID>` 
```

README.md
```diff
@@ -59,7 +59,8 @@ your own repository.
 
 1. Login to CodeClimate and add your project
 2. Go to `https://codeclimate.com/repos/<your-project>/settings/test_reporter`
-3. Copy the test reporter ID to `CODE_CLIMATE_TEST_REPORTER_TOKEN` in `app.json`
+3. Set up a Heroku pipeline.  
+4. `heroku config:set CC_TEST_REPORTER_ID=<Code Climate test reporter ID>` 
 4. Update https://codeclimate.com/repos/REPO_ID/settings/test_reporter to be at 100% diff coverage
```
prrrnd:
```suggestion
5. Update https://codeclimate.com/repos/REPO_ID/settings/test_reporter to be at 100% diff coverage
```

PR #31, opened 2020-01-08
Configured so ordering are using created_at

DeWinther:
When using UUID's, the active record methods 'first' and 'last' doesn't
work as expected, given that it's expecting an auto-incremented
id to sort by.


prrrnd:
👍

PR #33, opened 2020-01-13
Set up SendGrid by default

prrrnd:
https://app.asana.com/0/1142794766483633/1156764737042459

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -16,6 +16,7 @@ Bundler/OrderedGems:
 Metrics/BlockLength:
   Exclude:
     - "spec/**/*_spec.rb"
+    - "config/environments/**/*.rb"
 
 Metrics/LineLength:
   Max: 100
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -64,6 +64,20 @@ Rails.application.configure do
 
   config.action_mailer.perform_caching = false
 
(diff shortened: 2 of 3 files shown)


prrrnd:
@hcarreras The resource is added automatically based on the addons array in `app.json` :+1:

.rubocop.yml
```diff
@@ -16,6 +16,7 @@ Bundler/OrderedGems:
 Metrics/BlockLength:
   Exclude:
     - "spec/**/*_spec.rb"
+    - "config/environments/**/*.rb"
```
runephilosof-abtion: Maybe the mailer config should have been placed in its own initializer...

config/environments/production.rb
```diff
@@ -64,6 +64,20 @@
...
   config.action_mailer.perform_caching = false
 
+  config.action_mailer.delivery_method = :smtp
+
+  config.action_mailer.perform_deliveries = true
+
+  config.action_mailer.smtp_settings = {
+    user_name: ENV["SENDGRID_USERNAME"],
+    password: ENV["SENDGRID_PASSWORD"],
+    domain: "PROJECT_DOMAIN",
```
runephilosof-abtion: As I recall, domain is an optional setting.
If set, it should be set with a `ENV` since it will be different in the various environments.

PR #35, opened 2020-01-13
Set up rubocop-rails

prrrnd:
https://app.asana.com/0/1142794766483633/1156798276027955

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,6 @@
 require:
   - rubocop-performance
+  - rubocop-rails
 
 AllCops:
   TargetRubyVersion: 2.6
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -29,6 +29,9 @@ Metrics/MethodLength:
   Exclude:
     - "db/migrate/**/*.rb"
 
+Rails/FilePath:
+  EnforcedStyle: arguments
+
 Style/FrozenStringLiteralComment:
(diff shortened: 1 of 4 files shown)


prrrnd:
I'd love some input on the failing checks :-)

bohme:
it complains about making the path windows friendly 🤔 
I think we should keep paths windows friendly.

and it doesn;t like non-null columns without a default value... mhhhh... I am not sure I agree with that... anyone smarter than me that has thought harder about that one?

runephilosof-abtion:
> it complains about making the path windows friendly thinking
> I think we should keep paths windows friendly.

I think Windows handle forward slashes in paths quite nicely nowadays...?
Do we have any devs running Windows in Abtion?

> and it doesn;t like non-null columns without a default value... mhhhh... I am not sure I agree with that... anyone smarter than me that has thought harder about that one?

It is because the migration will fail if the table already contains rows.
If you make a new table with a non-null column, rubocop-rails won't complain about the missing default.

So if you really want the column to not have a default, you would need to either drop and recreate the table, or add the column with a default and the alter it to remove the default.

prrrnd:
I merged the migrations into one file, preventing us having to set a not null constraints on a populated db table.

I set the Rails/FilePath style to 'arguments', which is the style we have been using so far.

Ref: https://docs.rubocop.org/projects/rails/en/stable/cops_rails/#railsfilepath

PR #36, opened 2020-01-13
Prevent timing attacks in API authentication

prrrnd:
By first finding the user by its ID, and then comparing the
authentication tokens, we prevent timing attacks.
This solution however requires sending the user's ID alongside its
token.

### DIFF
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -10,14 +10,18 @@ module Api
       def require_authentication!
         user = User.find(request.headers["X-User-Id"])
 
-        if user.present? && Devise.secure_compare(
-          "Token #{user.authentication_token}",
-          request.headers["Authorization"]
-        )
+        if user.present? && valid_token?(user.authentication_token)
           sign_in user
         else
           render status: :unauthorized
         end
+      end
+
+      def valid_token?(token)
+        ActiveSupport::SecurityUtils.secure_compare(
+          "Token #{token}",
+          request.headers["Authorization"]
+        )
       end
     end
   end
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/base_controller.rb
```diff
@@ -8,11 +8,12 @@ class BaseController < ApplicationController
...
-        user = authenticate_or_request_with_http_token do |token, _options|
-          User.find_by(authentication_token: token)
-        end
+        user = User.find(request.headers["X-User-Id"])
 
-        if user.present?
+        if user.present? && Devise.secure_compare(
+          "Token #{user.authentication_token}",
+          request.headers["Authorization"]
+        )
```
bohme: could this one be extracted to gain a nicer name?
like... .`token_valid?` or the like?
I am not a fan of stretching an if over multiple lines 🤔

hcarreras:
I don't understand the vulnerability. Could you explain it better or share the source?

heatherm:
> I don't understand the vulnerability. Could you explain it in more depth or share the source?

More info here. https://app.asana.com/0/1142794766483633/1156142856809501/f

Perhaps we should use aid for PRing here so that the asana task gets linked?

app/controllers/api/v1/base_controller.rb
```diff
@@ -8,11 +8,12 @@ class BaseController < ApplicationController
       private
 
       def require_authentication!
-        user = authenticate_or_request_with_http_token do |token, _options|
-          User.find_by(authentication_token: token)
-        end
+        user = User.find(request.headers["X-User-Id"])
 
-        if user.present?
+        if user.present? && Devise.secure_compare(
```
runephilosof-abtion:
```suggestion
        if user.present? && ActiveSupport::SecurityUtils.secure_compare(
```

PR #37, opened 2020-01-13
Update CI configuration for cross browser testing

prrrnd:
https://app.asana.com/0/1142794766483633/1154327313115401

### DIFF
--- a/script/ci
+++ b/script/ci
@@ -14,7 +14,8 @@ export GIT_COMMITTED_AT="$(date +%s)"
 bundle exec rubocop
 
 # run the ruby test suite
-bundle exec rake spec
+CAPYBARA_DRIVER=headless_chrome bundle exec rake spec
+CAPYBARA_DRIVER=headless_firefox bundle exec rake spec
 
 # send report to code climate
 ./cc-test-reporter after-build --exit-code $?
--- a/script/ci
+++ b/script/ci
@@ -10,12 +10,12 @@ export GIT_BRANCH="$HEROKU_TEST_RUN_BRANCH"
 export GIT_COMMIT_SHA="$HEROKU_TEST_RUN_COMMIT_VERSION"
 export GIT_COMMITTED_AT="$(date +%s)"
 
-# run linter
-bundle exec rubocop
-
 # run the ruby test suite
 CAPYBARA_DRIVER=headless_chrome bundle exec rake spec
 CAPYBARA_DRIVER=headless_firefox bundle exec rake spec
+
+# run linter
+bundle exec rubocop
 
 # send report to code climate
 ./cc-test-reporter after-build --exit-code $?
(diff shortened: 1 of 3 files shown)


script/ci
```diff
@@ -10,11 +10,12 @@ export GIT_BRANCH="$HEROKU_TEST_RUN_BRANCH"
 export GIT_COMMIT_SHA="$HEROKU_TEST_RUN_COMMIT_VERSION"
 export GIT_COMMITTED_AT="$(date +%s)"
 
+# run the ruby test suite
+CAPYBARA_DRIVER=headless_chrome bundle exec rake spec
+CAPYBARA_DRIVER=headless_firefox bundle exec rake spec
```
runephilosof-abtion: Maybe only run the non-feature specs once. Even though they are fast.
```
bundle exec rspec --tag ~type:feature
CAPYBARA_DRIVER=headless_chrome bundle exec rspec --tag type:feature
CAPYBARA_DRIVER=headless_firefox bundle exec rspec --tag type:feature
```

prrrnd: Good point with code coverage. This is very much work in progress, and I just wanted to play around with Heroku CI and buildpacks. I'll have to get back to this next week.

GeoffAbtion:
What are we trying to test here?

runephilosof-abtion:
If the developers are only using Chrome they will not catch stuff not working in Firefox.
For instance some use datetime-local, expecting it to work in all browsers.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local

prrrnd:
@GeoffAbtion We are bound to supporting the latest version of all major browsers for most of our projects, so we should also reflect that in our test suites.

heatherm:
We _are_ bound to supporting the latest browsers via most SLAs, so I understand the intent of this PR. That said, I do think we need to find a better way (long term) to ensure cross browser functionality without doubling test time for e2es. I could see:

1. outsource to something like saucelabs/browserstack
2. randomize the browser selected for each test run
3. parallelize the tests w knapsack or something
4. update the SLAs to only specify one browser

Thoughts on ideal way to balance those tensions?

runephilosof-abtion:
For the CI it would be easy to keep the test time unchanged by parallelize the tests on separate test runs
https://devcenter.heroku.com/articles/heroku-ci-parallel-test-runs

Just do something like
if [[ "$CI_NODE_INDEX" == "0" ]]; then
  CAPYBARA_DRIVER=headless_chrome bundle exec rake spec
fi
if [[ "$CI_NODE_INDEX" == "1" ]]; then
  CAPYBARA_DRIVER=headless_firefox bundle exec rake spec
fi
etc.

Then for non CI test runs we could choose to only run the Firefox ones.
I chose Firefox since most of our developers use Chrome (although, I use Firefox), so it is most probable to make stuff incompatible with another browser than Chrome.

prrrnd:
Closing this one for now, as priorities have shifted.

PR #39, opened 2020-01-16
Fix structure.sql

prrrnd:


runephilosof-abtion:
I am assuming this change occurs because the Postgres version has changed.
Which Postgres version makes this change?
Newly created apps in Heroku uses Postgres 11.x (Muffi-template is using 11.6). So you shouldn't upgrade past that in your development environment.

prrrnd:
Yes, I had a newer Postgres version, and wrongly committed changes related to it. This reverts it.

runephilosof-abtion:
> This reverts it.

This? I see neither PR nor any commits.

runephilosof-abtion:
Oh, I thought this PR was created because you upgraded to Postgres 12, but this PR was the fix because you downgraded to the correct version.

PR #40, opened 2020-02-05
Adding a few cops 

heatherm:
This PR fixes the remaining rubocop errors and adds cops that have come up in pairing this week as useful additions.

The exclusion of application.rb as a whole from Debugger and ClassAndModuleCamelCase replaces a prior style of disable/enable per line. I wound up doing this as the code formatting was a nightmare on the line by line trying to get the legal `console` in the file to be ignored.


.rubocop.yml
```diff
@@ -17,18 +21,43 @@ Bundler/DuplicatedGem:
...
+  Exclude:
+    - "spec/**/*"
+
 Metrics/BlockLength:
+  Max: 50
   Exclude:
-    - "spec/**/*_spec.rb"
-    - "config/environments/**/*.rb"
+    - "config/**/*"
+    - "spec/**/*"
```
prrrnd: Did you run into anything specific here that requires ignoring more than just the *_spec.rb files?

.rubocop.yml
```diff
@@ -17,18 +21,43 @@ Bundler/DuplicatedGem:
...
 
-Metrics/LineLength:
-  Max: 100
+Metrics/ClassLength:
+  Max: 250
```
prrrnd: This seems like a lot of lines for a class? The default is already at 100.

config/application.rb
```diff
@@ -21,10 +21,7 @@
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
-
-# rubocop:disable Naming/ClassAndModuleCamelCase
```
prrrnd: I don't agree with this change and the addition to the .rubocop.yml file. When renaming the project, the lines here were close to the problem. By moving them to .rubocop.yml, we practically guarantee that the rule won't be removed when the problem is fixed.

heatherm:
Cool, feel free to edit and commit with what you find additive!

.rubocop.yml
bohme: yea I disagree with this change as well...
if you class has that many lines it might be doing too much 📟

.rubocop.yml
bohme: same for the configs... shouldn't they also contain readable code?

spec/rails_helper.rb
```diff
@@ -41,7 +41,7 @@
   exit 1
 end
 
-Dir.glob("./spec/support/**/*.rb").each { |f| require f }
+Dir.glob("./spec/support/**/*.rb").sort.each { |f| require f }
```
bohme: why does it need sorting? 🤔

.rubocop.yml
```diff
@@ -17,18 +21,43 @@ Bundler/DuplicatedGem:
...
 
 Metrics/MethodLength:
+  Max: 25
```
bohme: WOA! :O
I thought we wanted readable code?
I vote `10`

.rubocop.yml
```diff
@@ -17,18 +21,43 @@ Bundler/DuplicatedGem:
...
+  Exclude:
+    - "Rakefile"
+
+Metrics/AbcSize:
+  Max: 35
+  Exclude:
+    - "spec/**/*"
+
 Metrics/BlockLength:
+  Max: 50
```
bohme: No :O

.rubocop.yml
```diff
@@ -17,18 +21,43 @@ Bundler/DuplicatedGem:
...
+Lint/Debugger:
+  Exclude:
+    - "config/application.rb"
+
+Lint/ScriptPermission:
+  Exclude:
+    - "Rakefile"
+
+Metrics/AbcSize:
+  Max: 35
```
bohme: isn't the default 15?
why should our methods be so complicated?

spec/rails_helper.rb
bohme: `Lint/NonDeterministicRequireOrder`... the more you know 💃

PR #51, opened 2020-04-20
Feature: Test that authentication fails gracefully

bohme:
🌊 🦑 :shipit:

### DIFF
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -13,7 +13,7 @@ module Api
         if user.present? && valid_token?(user.authentication_token)
           sign_in user
         else
-          render json: {}, status: :unauthorized
+          head :unauthorized
         end
       end
 
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -9,25 +9,22 @@ module Api
 
       def require_authentication!
         id = request.headers["X-User-Id"]
+        token = request.headers.fetch("Authorization", "")
 
         user = User.find_by(id: id)
 
-        if user.present? && valid_token?(user.authentication_token)
+        if user.present? && valid_token?(user.authentication_token, token)
           sign_in user
         else
           head :unauthorized
         end
       end
 
-      def valid_token?(token)
+      def valid_token?(user_token, provided_token)
         ActiveSupport::SecurityUtils.secure_compare(
-          "Token #{token}",
-          token_from_headers
+          "Token #{user_token}",
+          provided_token
         )
-      end
-
-      def token_from_headers
-        request.headers.fetch("Authorization", "")
       end
     end
   end
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -8,31 +8,12 @@ module Api
       private
 
       def require_authentication!
-        authenticator = Authenticator.new(
-          user: User.find_by(id: request.headers["X-User-Id"])
-        )
+        user = User.find_by(id: request.headers["X-User-Id"])
 
-        if authenticator.valid_token?(request.headers["Authorization"])
-          sign_in authenticator.user
+        if user&.valid_token?(request.headers["Authorization"])
+          sign_in user
         else
           head :unauthorized
-        end
-      end
-
-      class Authenticator
-        attr_reader :user
-
-        def initialize(user:)
-          @user = user
-        end
-
-        def valid_token?(provided_token)
-          return false unless user && provided_token
-
-          ActiveSupport::SecurityUtils.secure_compare(
-            "Token #{user.authentication_token}",
-            provided_token
-          )
         end
       end
     end
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -22,8 +22,6 @@ module Api
       end
 
       def valid_token?(token)
-        return unless token && token_from_headers
-
         ActiveSupport::SecurityUtils.secure_compare(
           "Token #{token}",
           token_from_headers
@@ -35,7 +33,7 @@ module Api
       end
 
       def token_from_headers
-        request.headers["Authorization"]
+        request.headers.fetch("Authorization", "")
       end
     end
   end
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -9,11 +9,10 @@ module Api
 
       def require_authentication!
         authenticator = Authenticator.new(
-          user: User.find_by(id: request.headers["X-User-Id"]),
-          provided_token: request.headers["Authorization"]
+          user: User.find_by(id: request.headers["X-User-Id"])
         )
 
-        if authenticator.valid?
+        if authenticator.valid_token?(request.headers["Authorization"])
           sign_in authenticator.user
         else
           head :unauthorized
@@ -21,14 +20,13 @@ module Api
       end
 
       class Authenticator
-        attr_reader :user, :provided_token
+        attr_reader :user
 
-        def initialize(user:, provided_token:)
+        def initialize(user:)
           @user = user
-          @provided_token = provided_token
         end
 
-        def valid?
+        def valid_token?(provided_token)
           return false unless user && provided_token
 
           ActiveSupport::SecurityUtils.secure_compare(
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -8,20 +8,34 @@ module Api
       private
 
       def require_authentication!
-        user = User.find(request.headers["X-User-Id"])
+        user = find_user
 
         if user.present? && valid_token?(user.authentication_token)
           sign_in user
         else
-          render status: :unauthorized
+          render json: {}, status: :unauthorized
         end
       end
 
+      def find_user
+        User.find_by(id: user_id_from_headers)
+      end
+
       def valid_token?(token)
+        return unless token && token_from_headers
+
         ActiveSupport::SecurityUtils.secure_compare(
           "Token #{token}",
-          request.headers["Authorization"]
+          token_from_headers
         )
+      end
+
+      def user_id_from_headers
+        request.headers["X-User-Id"]
+      end
+
+      def token_from_headers
+        request.headers["Authorization"]
       end
     end
   end
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -8,34 +8,20 @@ module Api
       private
 
       def require_authentication!
-        authenticator = Authenticator.new(
-          user: User.find_by(id: request.headers["X-User-Id"]),
-          provided_token: request.headers["Authorization"]
-        )
+        user = User.find_by(id: request.headers["X-User-Id"])
 
-        if authenticator.valid?
-          sign_in authenticator.user
+        if user.present? && valid_token?(user.authentication_token)
+          sign_in user
         else
           head :unauthorized
         end
       end
 
-      class Authenticator
-        attr_reader :user, :provided_token
-
-        def initialize(user:, provided_token:)
-          @user = user
-          @provided_token = provided_token
-        end
-
-        def valid?
-          return false unless user && provided_token
-
-          ActiveSupport::SecurityUtils.secure_compare(
-            "Token #{user.authentication_token}",
-            provided_token
-          )
-        end
+      def valid_token?(token)
+        ActiveSupport::SecurityUtils.secure_compare(
+          "Token #{token}",
+          request.headers.fetch("Authorization", "")
+        )
       end
     end
   end
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -8,23 +8,34 @@ module Api
       private
 
       def require_authentication!
-        id = request.headers["X-User-Id"]
-        token = request.headers.fetch("Authorization", "")
+        authenticator = Authenticator.new(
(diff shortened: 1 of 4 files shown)


spec/controllers/api/v1/base_controller_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        get :index
+
+        expect(response.status).to eq(401)
+      end
+    end
+
+    context "when no valid token and user id given" do
+      it "returns a 401 unauthorized" do
+        request.headers["X-User-Id"] = "nope"
+        request.headers["Authorization"] = "Token 'nope'"
```
runephilosof-abtion, line 75:
```suggestion
```

bohme: uhh! thanks :3

app/controllers/api/v1/base_controller.rb
```diff
@@ -8,21 +8,35 @@ class BaseController < ApplicationController
...
+          render json: {}, status: :unauthorized
         end
       end
 
+      def find_user
+        User.find_by(id: user_id_from_headers)
+      end
+
       def valid_token?(token)
+        return unless token && token_from_headers
```
GeoffAbtion, line 25: This local / global argument mismatch bugs me.

app/controllers/api/v1/base_controller.rb
```diff
@@ -8,21 +8,35 @@ class BaseController < ApplicationController
...
 
       def require_authentication!
-        user = User.find(request.headers["X-User-Id"])
+        user = find_user
 
         if user.present? && valid_token?(user.authentication_token)
           sign_in user
         else
-          render status: :unauthorized
+          render json: {}, status: :unauthorized
```
GeoffAbtion, line 16: Do you need to render anything here?
You're not testing for this.
Maybe use `header` instead of `render`

spec/controllers/api/v1/base_controller_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+        expect(response.status).to eq(401)
+      end
+    end
+
+    context "when no valid token and user id given" do
+      it "returns a 401 unauthorized" do
+        get :index
+
+        expect(response.status).to eq(401)
```
GeoffAbtion, line 76: Can we do this: 
`expect(response).to have_http_status(:unathorized)`

GeoffAbtion: Is it an improvement?

app/controllers/api/v1/base_controller.rb
bohme: uhh! `header` is perfect for this :3

spec/controllers/api/v1/base_controller_spec.rb
bohme: does make it more readable, arguably the numbers are for computers and the words for humans 💃

app/controllers/api/v1/base_controller.rb
bohme: what exactly do you mean? 🤔

spec/controllers/api/v1/base_controller_spec.rb
GeoffAbtion: I'm not sure it is an improvement, since it adds a bunch of syntactic sugar over what was previously a straightforward assertion. And now you need to remember that 401 is `:unauthorized` when you're comparing numeric values returned by your test failures.

But it is more rspec-like

heatherabtion:
🚢

app/controllers/api/v1/base_controller.rb
```diff
@@ -8,21 +8,35 @@ class BaseController < ApplicationController
       private
 
       def require_authentication!
-        user = User.find(request.headers["X-User-Id"])
+        user = find_user
 
         if user.present? && valid_token?(user.authentication_token)
           sign_in user
```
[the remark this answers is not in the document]
GeoffAbtion: (it is now)

app/controllers/api/v1/base_controller.rb
```diff
@@ -8,20 +8,34 @@ class BaseController < ApplicationController
...
+      class Authenticator
+        attr_reader :user, :provided_token
+
+        def initialize(user:, provided_token:)
+          @user = user
+          @provided_token = provided_token
+        end
+
+        def valid?
+          return false unless user && provided_token
```
GeoffAbtion, line 32: following on from: https://github.com/abtion/muffi/pull/51#discussion_r415589660

I was a bit surprised that this line ended up in the refactored version too:
`          return false unless user && provided_token`

But here it doesn't bug me as much.

It's still a guard clause, but it's guarding against 2 things that are in the same scope, which is as arguments passed into our constructor. 
They are 2 things which are fundamental to the authentication, that there is a user and that there is a provided token that we need to compare to.

I used `provided_token` here to more clearly differentiate from the `authentication_token` which is a property of the user model.

app/controllers/api/v1/base_controller.rb
```diff
@@ -8,34 +8,20 @@ class BaseController < ApplicationController
...
-          ActiveSupport::SecurityUtils.secure_compare(
-            "Token #{user.authentication_token}",
-            provided_token
-          )
-        end
+      def valid_token?(token)
+        ActiveSupport::SecurityUtils.secure_compare(
+          "Token #{token}",
+          request.headers.fetch("Authorization", "")
+        )
```
GeoffAbtion, line 24: Now that we have all those tests we can refactor away the Authenticator class to leave us with an arguably simpler implementation that doesn't change so much of the original code.

I have a slight preference for the `Authenticator` version. 
But I wanted to show that this version was also possible.

GeoffAbtion:
Look at https://github.com/abtion/muffi/pull/51/files/f0ce8ac29729bc8e849d36da82edaff1cfab1bc8 for the "Simpler" version that does not use the Authenticator class.

It is less code, but has a conditional that does a lot of things and the global `request.headers` get accessed in a lot of different places, where the class version neatly encapsulates everything.

GeoffAbtion:
🚲 🏠

app/controllers/api/v1/base_controller.rb
bohme: You make a very clear point, I like it 👍

app/controllers/api/v1/base_controller.rb
```diff
@@ -8,20 +8,34 @@ class BaseController < ApplicationController
...
 
-      def valid_token?(token)
-        ActiveSupport::SecurityUtils.secure_compare(
-          "Token #{token}",
-          request.headers["Authorization"]
-        )
+      class Authenticator
+        attr_reader :user, :provided_token
+
+        def initialize(user:, provided_token:)
```
bohme, line 26: would renaming `provided_token` to `token` make this an ever so slightly better name 🤔 
The attribute being prefixed with `provided` makes sense to indicate where it comes from, but it bothers me a little that the name of the parameter has the prefix 🤔

GeoffAbtion: I always want to be explicit about the token that we've been provided by the person trying to log in and any token that might be valid for the user. Just calling it `token` introduces some ambiguity.
If you see it in isolation:

`Authenticator.new(user: whoever, token: '123xyz')`

Is this going to be used for comparing? Is it assigning a token to be used for that user which will be validated later some other way?

`provided_token` name still bugs me a bit because it's still a bit unclear what "provided" might mean.

All this is probably a sign that there's some refactoring hiding in here that could make things clearer.

bohme: Can similar questions be raised about `user`?
To me it appears relatively clear that the `Authenticator`  will try to authenticate the `user` with said `token` 🤔 
I see your argument though... the `Authenticator` could assign the `token` to the `user` if the `user` is valid, name wise.

sneaky one 👍 
thanks ^__^

app/controllers/api/v1/base_controller.rb
```diff
@@ -8,33 +8,14 @@ class BaseController < ApplicationController
...
           head :unauthorized
         end
       end
-
-      class Authenticator
-        attr_reader :user
-
-        def initialize(user:)
-          @user = user
-        end
```
GeoffAbtion: When a class only takes one constructor argument, that's a good sign that the method(s) should be on that class.
(Unless your class really should be a decorator, which I don't think is true here.)

app/controllers/api/v1/base_controller.rb
```diff
@@ -8,33 +8,14 @@ class BaseController < ApplicationController
...
 
       def require_authentication!
-        authenticator = Authenticator.new(
-          user: User.find_by(id: request.headers["X-User-Id"])
-        )
+        user = User.find_by(id: request.headers["X-User-Id"])
 
-        if authenticator.valid_token?(request.headers["Authorization"])
-          sign_in authenticator.user
+        if user&.valid_token?(request.headers["Authorization"])
```
GeoffAbtion, line 13: `&.` 😿

spec/models/concerns/api_authenticable_spec.rb
```diff
@@ -12,4 +12,23 @@
...
+  describe "valid_token?" do
+    it "Securely compares the tokens (prefixed with 'Token ')" do
+      user = build(:user, authentication_token: "SecurelyRandomToken")
+
+      allow(ActiveSupport::SecurityUtils).to receive(:secure_compare)
+        .and_return(:secure_compare_boolean_result)
+
+      expect(user.valid_token?("Token SecurelyRandomToken")).to eq :secure_compare_boolean_result
+      expect(ActiveSupport::SecurityUtils).to have_received(:secure_compare)
+        .with("Token SecurelyRandomToken", "Token SecurelyRandomToken")
```
bohme, line 25: testing that it returns the symbol feels weird, we don't care about that in this spec, what we care about is the token and how it compares:
```suggestion
      user = build(:user, authentication_token: "SecurelyRandomToken")
      token = "SecurelyRandomToken"

      allow(ActiveSupport::SecurityUtils).to receive(:secure_compare)

      user.valid_token?("Token #{token}")
      expect(ActiveSupport::SecurityUtils).to have_received(:secure_compare)
        .with("Token SecurelyRandomToken", "Token SecurelyRandomToken")
```

app/controllers/api/v1/base_controller.rb
GeoffAbtion: > All this is probably a sign that there's some refactoring hiding in here that could make things clearer.

See the next 2 commits where I do this refactoring.

spec/models/concerns/api_authenticable_spec.rb
GeoffAbtion: But we do need to care that the result of the `secure_compare` is what is being returned, so I think it is necessary.

locally, with this change applied, try adding a `true` as the return value for `valid_token?` and see if any tests fail.

GeoffAbtion: I use `:secure_compare_boolean_result` as a stub return value here because we need to check that that value is being returned and we don't care in this test whether it is true or false.

We could write two tests, one for token matching and one for token not matching, but if we don't want to call the real method we need to fake the return value twice as well, and what's the point of doing it twice if you only need to do it once? 🤷

bohme: fair, actually testing that `ActiveSupport::SecureUtils` works feels silly 🤔

prrrnd:
While the work here is appreciated, I think it should go through Asana first and be "approved". This PR has 30 comments - What is blocking it now?

app/controllers/api/v1/base_controller.rb
bohme: gonna be tricky to get rid of this one 💃

app/models/concerns/api_authenticable.rb
```diff
@@ -16,5 +16,14 @@ def generate_authentication_token
...
     end
+
+    def valid_token?(provided_token)
+      return false unless provided_token
+
+      ActiveSupport::SecurityUtils.secure_compare(
+        "Token #{authentication_token}",
+        provided_token
+      )
+    end
```
bohme, line 27: this is quite elegant 💪

app/controllers/api/v1/base_controller.rb
GeoffAbtion: 😿 And until Rails adds proper null object handling, I'll keep leaving crying cat faces in PR comments.

bohme: Why not add a NullUser than :D

app/controllers/api/v1/base_controller.rb
```diff
@@ -8,14 +8,20 @@ class BaseController < ApplicationController
...
         else
           head :unauthorized
         end
       end
+
+      class NullUser
+        def valid_token?(_token)
+          false
+        end
+      end
```
GeoffAbtion, line 24: Yeah, this is not good.

(More explanation when I get time.)

GeoffAbtion:
We just merged a branch that fails CI 😱

GeoffAbtion:
Ok, it seems to "only" have been a problem that codeclimate had some problems connecting to github, but still. 😟

app/controllers/api/v1/base_controller.rb
GeoffAbtion: It's really tricky to do it nicely.

app/controllers/api/v1/base_controller.rb
bohme: hahaha I expected as much ^^
unfortunate timing got it merged, guess both Pierre and I thought Muffi when it came to lunch 🐕

bohme:
for some positive news:
The last commit didn't actually make it into the merge :D
https://github.com/abtion/muffi/commit/719b4b43501b3f50cc14a1fa09e56f90564dea72

PR #52, opened 2020-04-21
Feature: Extract code_coverage_helper into own file

bohme:
Asana-tasks: [Extract code coverage helper file.](https://app.asana.com/0/1142794766483633/1156767475231026/f) & [Exclude /vendor from code coverage metrics](https://app.asana.com/0/1142794766483633/1156767475231022/f)

### DIFF
--- a/spec/code_coverage_helper.rb
+++ b/spec/code_coverage_helper.rb
@@ -5,5 +5,6 @@ require "simplecov"
 unless ENV["DISABLE_SIMPLECOV"] == "true"
   SimpleCov.start("rails") do
     add_filter "spec"
+    add_filter "vendor"
   end
 end
--- a/spec/code_coverage_helper.rb
+++ b/spec/code_coverage_helper.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+require "simplecov"
+
+unless ENV["DISABLE_SIMPLECOV"] == "true"
+  SimpleCov.start("rails") do
+    add_filter "spec"
+  end
+end
--- a/spec/code_coverage_helper.rb
+++ b/spec/code_coverage_helper.rb
@@ -2,4 +2,9 @@
 
 require "simplecov"
 
-SimpleCov.start("rails") unless ENV["DISABLE_SIMPLECOV"] == "true"
+unless ENV["DISABLE_SIMPLECOV"] == "true"
+  SimpleCov.start("rails") do
+    add_filter "spec"
+    add_filter "vendor"
+  end
+end
--- a/spec/code_coverage_helper.rb
+++ b/spec/code_coverage_helper.rb
@@ -2,8 +2,4 @@
 
 require "simplecov"
 
-unless ENV["DISABLE_SIMPLECOV"] == "true"
-  SimpleCov.start("rails") do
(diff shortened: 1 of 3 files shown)


spec/code_coverage_helper.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require "simplecov"
+
+unless ENV["DISABLE_SIMPLECOV"] == "true"
+  SimpleCov.start("rails") do
+    add_filter "vendor"
```
heatherabtion, line 7: filter is updated to `vendor` from `spec`. Intentional or?

bohme: someone didn't read the PR-message 🤔

bohme: ah.. forgot to read spec, sneaky one 🤔

bohme: mhh... after a little bit of research I don't think we need to exclude the `spec` folder.
The generated report is identical with or without the filter.

runephilosof-abtion: Simplecov adds `spec` and `vendor` for us. https://github.com/colszowka/simplecov/blob/master/lib/simplecov/profiles/test_frameworks.rb https://github.com/colszowka/simplecov/blob/4e405d5fe0520a43f6757039444a423fed928c6a/lib/simplecov/profiles/bundler_filter.rb
`vendor` is added by bundler_filter which is added by default. https://github.com/colszowka/simplecov/blob/4e405d5fe0520a43f6757039444a423fed928c6a/lib/simplecov/defaults.rb#L15
`spec` is added by test_framework which is added by the rails profile. https://github.com/colszowka/simplecov/blob/4e405d5fe0520a43f6757039444a423fed928c6a/lib/simplecov/profiles/rails.rb

```suggestion
```

bohme: uhh nice research 💪

heatherabtion: My experience is that these are not respected in the version of simplecov used by codeclimate.

.env.test
```diff
@@ -1,3 +1,4 @@
 SEED_ADMIN_EMAIL=admin@example.com
 SEED_ADMIN_INITIAL_PASSWORD=password
+CODE_COVERAGE_PERCENTAGE=70
```
heatherabtion, line 3: How did we arrive at this number?

bohme: Would like it to be `100%` but current coverage when merging this would be `73.56%`.
So I took the current coverage and rounded down to the nearest pretty number.
Do you have a different number in mind?

GeoffAbtion: 73

bohme: but 73 is an ugly number 🤔

GeoffAbtion: 100 is the only good number 🤷

prrrnd: Could we separate the two? This PR for moving the existing code coverage helper to its own file, and an other PR to raise code coverage and set the minimum?

bohme: sure ^__^

bohme: Extracted to here: https://github.com/abtion/muffi/pull/54

PR #53, opened 2020-04-27
Fix: Unbreak yarn

bohme:
Run your code folks!

By locking 'kind-of' to a high version number we were ignoring all the
past major versions. Removing them from the 'yarn.lock' broke our
JS-dependencies.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3530,6 +3530,11 @@ is-binary-path@^1.0.0:
   dependencies:
     binary-extensions "^1.0.0"
 
+is-buffer@^1.1.5:
+  version "1.1.6"
+  resolved "https://registry.yarnpkg.com/is-buffer/-/is-buffer-1.1.6.tgz#efaa2ea9daa0d7ab2ea13a97b2b8ad51fefbe8be"
+  integrity sha512-NcdALwpXkTm5Zvvbk7owOUSvVvBKDgKP5/ewfXEznmQFfs4ZRmanOeKBTjRVjka3QFoN6XJ+9F3USqfHqTaU5w==
+
 is-callable@^1.1.4, is-callable@^1.1.5:
   version "1.1.5"
   resolved "https://registry.yarnpkg.com/is-callable/-/is-callable-1.1.5.tgz#f7e46b596890456db74e7f6e976cb3273d06faab"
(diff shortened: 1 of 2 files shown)


GeoffAbtion:
How does this interact with the recent `kind-of` security issues?

GeoffAbtion:
This fixes the issues I was having though 👍

GeoffAbtion:
🤔 Why did CI not fail before this fix?

bohme:
Oh yea fair point:
The issue was in `6.0.2` we have the `^6.0.0` requirement at `6.0.3` now which fixes the issue https://github.com/jonschlinkert/kind-of/issues/30

CI didn't fail because the specs run fine 🦐

runephilosof-abtion:
> This fixes the issues I was having though +1

What issues?
I just upgraded another project by setting resolutions of kind-of to `^6.0.3`...

bohme:
Good old:
```
13:47:10 bundle exec rails s
=> Booting Puma
=> Rails 6.0.2.2 application starting in development
=> Run `rails server --help` for more startup options


========================================
  Your Yarn packages are out of date!
  Please run `yarn install --check-files` to update.
========================================


To disable this check, please change `check_yarn_integrity`
to `false` in your webpacker config file (config/webpacker.yml).


yarn check v1.22.4
success Folder in sync.
Done in 0.08s.
yarn check v1.22.4
error "webpack-dev-server#chokidar#braces#fill-range#is-number#kind-of" is wrong version: expected "^3.0.2", got "6.0.3"
error "webpack-dev-server#chokidar#braces#snapdragon-node#snapdragon-util#kind-of" is wrong version: expected "^3.2.0", got "6.0.3"
error "@rails/webpacker#webpack#micromatch#snapdragon#define-property#is-descriptor#kind-of" is wrong version: expected "^5.0.0", got "6.0.3"
error "@rails/webpacker#webpack#micromatch#snapdragon#base#cache-base#has-value#has-values#kind-of" is wrong version: expected "^4.0.0", got "6.0.3"
error "@rails/webpacker#webpack#micromatch#snapdragon#base#class-utils#static-extend#object-copy#kind-of" is wrong version: expected "^3.0.3", got "6.0.3"
error Found 5 errors.
info Visit https://yarnpkg.com/en/docs/cli/check for documentation about this command.


Exiting
```

bohme:
I do like how webpack needs type-of version 3, 4 and 5 🤔

bohme:
and yes, I had run `yarn install --check-files` before 🌊

runephilosof-abtion:
So it is "only" a problem in development.
Nothing has broken in production?

runephilosof-abtion:
I did this resolutions upgrade in another project with no problems.
Same version of Rails.

bohme:
you got a link to the commit/yarn.lock ?

yarn.lock
```diff
@@ -3858,7 +3863,26 @@ killable@^1.0.1:
...
+  dependencies:
+    is-buffer "^1.1.5"
+
+kind-of@^5.0.0:
+  version "5.1.0"
+  resolved "https://registry.yarnpkg.com/kind-of/-/kind-of-5.1.0.tgz#729c91e2d857b7a419a1f9aa65685c4c33f5845d"
+  integrity sha512-NGEErnH6F2vUuXDh+OlbcKW7/wOcfdRHaZ7VWtqCztfHri/++YKmP51OdWeGPuqCOba6kk2OTe5d02VmTB80Pw==
+
+kind-of@^6.0.0, kind-of@^6.0.2:
   version "6.0.3"
```
runephilosof-abtion, line 3886: The important part is that this says 6.0.3 and not 6.0.2.
We don't need to persist that version decision in package.json.
We have a system in place (dependabot) that complains if we for some reason ever downgrade to 6.0.2.
So we can remove the resolutions part entirely in package.json.

bohme: uhh someone looking at Muffi PRs <3

removed the resolution part, deleted `node_modules`, reinstalled yarn packages and commit everything.

What do ya think? ^__^

runephilosof-abtion: That would be fine.

We have just gone through the exercise of updating all dependencies in a new project based on Muffi.
Also, we set up Dependabot to keep it updated (also for non-security updates).
Will hopefully soon be adding a PR in Muffi with those changes.

bohme: so would you say you'd approve of it now? ^___^

runephilosof-abtion: yes

bohme:
weeeeee :3

PR #55, opened 2020-04-29
Feature: Add empty spec files to please SimpleCov

bohme:
Adding empty files to test empty files.

Hunting those 100% 🤠

### DIFF
--- a/spec/channels/application_cable/channel_spec.rb
+++ b/spec/channels/application_cable/channel_spec.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe ApplicationCable::Channel do
+end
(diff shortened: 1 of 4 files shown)


GeoffAbtion:
Better solution, delete the empty files.

bohme:
We could do my favourite rails thingy and split the `rails` gem into it's components :P

runephilosof-abtion:
> Better solution, delete the empty files.

ActionCable could be deleted, since a lot of projects don't use it.
But jobs and mailers are typically used, we would want projects to inherit from ApplicationMailer as is Rails default.

GeoffAbtion:
> But jobs and mailers are typically used, we would want projects to inherit from ApplicationMailer as is Rails default.

At which point we could create the needed files.

GeoffAbtion:
We should either keep all of them or none of them, but either way there should be test files for the existing files.

I vote we merge this PR and we can argue about deleting things in a separate PR.

PR #56, opened 2020-04-29
Feature: Add specs for Ability

bohme:
Hunting those 100% 🤠


spec/models/ability_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+require "cancan/matchers"
+
+RSpec.describe Ability, type: :model do
+  context "when user is admin" do
```
bohme, line 7: but what when user is user 🌊 
and what if user is nothing at all :O

PR #57, opened 2020-04-29
Feature: Add specs for HomeController

bohme:
Hunting those 100% 🤠

### DIFF
--- a/spec/controllers/home_controller_spec.rb
+++ b/spec/controllers/home_controller_spec.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe HomeController, type: :controller do
+  describe "#index" do
+    it "renders 'abtion-rails'" do
+      user = FactoryBot.create(:user)
+      request.headers["X-User-Id"] = user.id
+      request.headers["Authorization"] = "Token #{user.authentication_token}"
+
+      get :index, params: { id: user.id }
+
+      expect(response.body).to eq("abtion-rails")
+    end
+  end
+end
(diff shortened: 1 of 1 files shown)


spec/controllers/home_controller_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+require "rails_helper"
+
+RSpec.describe HomeController, type: :controller do
+  describe "#index" do
+    it "renders 'abtion-rails'" do
+      user = FactoryBot.create(:user)
+      request.headers["X-User-Id"] = user.id
+      request.headers["Authorization"] = "Token #{user.authentication_token}"
+
+      get :index, params: { id: user.id }
```
GeoffAbtion, line 12: This is a horrible amount of setup to need to do.

spec/controllers/home_controller_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
+      expect(response.body).to eq("abtion-rails")
```
GeoffAbtion, line 14: Check the http_status too maybe?

spec/controllers/home_controller_spec.rb
GeoffAbtion: (it seems to be unnecessary?)

bohme: oh... no authentication needed for the `HomeController`  :O

spec/controllers/home_controller_spec.rb
bohme: but how can the body be correct if the status is anything but `200` 🤔 
ahhh... this will tell one the status code if it fails 👍

GeoffAbtion: `render status: 418, plain: "abtion-rails"`

spec/controllers/home_controller_spec.rb
GeoffAbtion: That may or may not be an issue. Either way it should be in a different PR.

spec/controllers/home_controller_spec.rb
bohme: think this is less of a teapot and more of a 218 ;)

PR #58, opened 2020-04-29
Feature: Add specs for ApplicationController#access_denied

bohme:
Our current implementation doesn't actually work and mine doesn't feel right 🤔


bohme:
1. The original code looks like it wants to render a page.
2. It feels silly to output the error message to the user.

Think I just wanna head the status?

GeoffAbtion:
The thing that should happen in this situation is very app-specific so I think what you're doing here is a reasonable default for a template to have.

PR #63, opened 2020-05-19
bin/setup: Create .env.development if it does not exist.

GeoffAbtion:
If the .env.development file does not exist, create it based on the
.env.development.example file.
If already exists, do nothing.

This allows an app to be setup from git clone to passing tests by a single: `bin/setup` command.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -25,6 +25,11 @@ FileUtils.chdir APP_ROOT do
   #   FileUtils.cp 'config/database.yml.sample', 'config/database.yml'
   # end
 
+  unless File.exist?(".env.development")
+    puts "\n== Creating .env.development from example file"
+    FileUtils.cp ".env.development.example", ".env.development"
+  end
+
   puts "\n== Preparing database =="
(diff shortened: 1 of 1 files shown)


GeoffAbtion:
I agree that that would be a better way of handling environment variables. 👍

jeppester:
> I prefer to not use `.example`.
> Instead track `.env` and `.env.development` in Git.
> Then it is even easier to get running locally, also it is the documented way of doing it in the `dotenv` gem => the expected way
> 
> Instead rename `.env.development.example` to `.env.development`.
> If developers need local changes to the development env variables, they can create `.env.development.local`
> Remove `.env.development` and `.env` from `.gitignore`.
> Add `.env.*.local` to `.gitignore`

This should be a simple change, lets get it done!

jeppester:
I'm closing this in favor of @runephilosof-abtion's solution, I'll make a PR for it.

GeoffAbtion:
Great, thanks @jeppester 👍

PR #74, opened 2020-08-05
Replace CodeClimate's Brakeman and use Bundler instead

Guldbek:
We won't need to worry about having 2 different versions of brakeman running.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -6,13 +6,6 @@ exclude_patterns:
   - public/**/*
 
 plugins:
-  # https://docs.codeclimate.com/docs/brakeman
-  brakeman:
-    enabled: true
-    checks:
-      no_attr_accessible:
-        enabled: false
-
   # https://docs.codeclimate.com/docs/bundler-audit
   bundler-audit:
     enabled: true
--- a/script/ci
+++ b/script/ci
@@ -16,5 +16,8 @@ bundle exec rubocop
 # run the ruby test suite
 bundle exec rake spec
 
+# run brakeman
+bundle exec brakeman
+
 # send report to code climate
 ./cc-test-reporter after-build --exit-code $?


jeppester:
The same could be said about rubocop, but code climate has better integration with github, and the more stuff we can get out of our CI and into code climate, the faster the CI will be.

So I'm not sure that I agree on making this change.
I'm ready to be convinced otherwise though. :slightly_smiling_face:

runephilosof-abtion:
The point is that Brakeman in CodeClimate is a very old version giving us false positives in new Rails.
Because of some licensing issues CodeClimate is not allowed to update their Brakeman version.
https://docs.codeclimate.com/docs/brakeman:
> Brakeman version 4.3.1
>     🚧       Due to the acquisition of Brakeman, we're currently not able to upgrade our OSS Brakeman plugin to use Brakeman's latest version.
The newest Brakeman version on rubygems https://rubygems.org/gems/brakeman/ is 4.9.0.

If we want to use Brakeman on CodeClimate we should buy a subscription https://www.brakemanpro.com/docs/codeclimate.

runephilosof-abtion:
The same cannot be said about Rubocop.

runephilosof-abtion:
But we should put some of that reasoning in the code and commit message.

jeppester:
@runephilosof-abtion Makes sense!, then lets make sure no-one will spend time on trying to move brakeman back into code climate then, just to find out what you just did.

heatherabtion:
It is fine to purchase a subscription to brakeman-pro if that fits the flow better for you all.

runephilosof-abtion:
In Fraxx, we have moved the other way. Now, we have also removed Rubocop from the CodeClimate config (because we run it in CI, in a newer version than CodeClimate supports).

jeppester:
If we want linters to run as part of our CI I think it would make sense to execute them in a parallel test job, so that they don't prolong the test suite.

The question is if it makes a lot of sense to work on CI now, I think there's a good chance that we'll move to another CI provider, since heroku CI is really slow when chrome buildpack is needed.

jeppester:
@bohme and I have implemented this as a part of https://github.com/abtion/muffi/pull/87

PR #83, opened 2020-08-25
Configure CSP header.

runephilosof-abtion:
https://app.asana.com/0/1142794766483633/1188627776772511
https://observatory.mozilla.org/analyze/muffi-feature-csp-heade-gbvrsq.herokuapp.com

This should give 120/100 score on https://observatory.mozilla.org.
It will probably give some surprise, when trying to link to a script hosted remotely,
but some browsers give good error messages, explaining why the script isn't loaded.

For instance, if using Stripe Checkout. You need to change (Stripe's documentation on this is currently wrong):
```
  policy.frame_src   "https://js.stripe.com"
  policy.script_src  :self, "https://js.stripe.com"
```

### DIFF
--- a/README.example.md
+++ b/README.example.md
@@ -1,5 +1,7 @@
 # Project Name Human
 
+![Observatory](https://img.shields.io/mozilla-observatory/grade-score/muffi-template.herokuapp.com)
+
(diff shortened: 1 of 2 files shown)


runephilosof-abtion:
Sorry, the PR isn't ready yet.

PR #85, opened 2020-08-25
Unknown method resource_name

runephilosof-abtion:
Asana task [Devise devise_mapping and resource_name](https://app.asana.com/0/1142794766483633/1187208358171569)


runephilosof-abtion:
This also contains an upgrade to rspec-rails because that is needed to support controller spec (with views) with Rails 6.

config/environments/development.rb
```diff
@@ -60,6 +60,8 @@
   # Raises error for missing translations.
   # config.action_view.raise_on_missing_translations = true
 
+  config.hosts << "dev-server"
```
bohme, line 63: 🤔 
where does this come from?

PR #86, opened 2020-08-27
Make it possible to locally add allowed hosts

jeppester:
I'm sorry about side-stepping the Muffi board.
This should be uncontroversial, and I need this for my remote dev env to work.


bohme (approved):
Nice one 👍

PR #88, opened 2020-08-27
Replace sprockets with webpacker

jeppester:
https://app.asana.com/0/1142794766483633/1187208358171573

Cherry-picked from fraxx:

* Remove sprockets.

To avoid having different versions of bootstrap in javascript and ruby.

* Remove app/assets

Move app/assets/fonts to app/javascript/fonts


spec/support/capybara.rb
```diff
@@ -34,3 +34,18 @@
...
+
+    if errors.present?
+      aggregate_failures "javascript errors/warnings" do
+        errors.each do |error|
+          raise error.message
+        end
+      end
+    end
+  end
+end
```
bohme, line 51: I love this and want it in every project 🥇

runephilosof-abtion: Or maybe this:
```
      aggregate_failures "javascript errors" do
        errors.each do |error|
          expect(error.level).not_to eq("SEVERE"), error.message
        end
      end
```

runephilosof-abtion: Often you can't avoid warnings

app/javascript/stylesheets/globals/_fonts.scss
```diff
@@ -1,11 +1,11 @@
...
-  src:font-url('D-DIN.otf');
+  src:url('../fonts/D-DIN.otf');
 }
 
 @font-face {
   font-family:'d-din-bold';
   font-weight: bold;
-  src:font-url('D-DIN-Bold.otf');
+  src:url('../fonts/D-DIN-Bold.otf');
 }
```
prrrnd, line 11: Needs a new line at EOL

prrrnd (commented):
What would be the flow for images? Perhaps it should be explained somewhere?

spec/support/capybara.rb
jeppester: We discussed that, but I'm not sure that I agree, I don't remember a lot of js warnings that should be ignored.
Maybe you have a recent example? :-)

app/javascript/stylesheets/globals/_fonts.scss
bohme: not part of this PR 🤷‍♀️

bohme:
```
// Uncomment to copy all static images under ../images to the output folder and reference
// them with the image_pack_tag helper in views (e.g <%= image_pack_tag 'rails.png' %>)
// or the `imagePath` JavaScript helper below.
//
// const images = require.context('../images', true)
// const imagePath = (name) => images(name, true)
```
will add a `app/javascript/images/.keep` for even easier use good catch @prrrnd 👍

heatherabtion (commented):
In `config/webpacker.yml` you can specify `source_path: app/frontend` to change the default name if you find it odd to nest non-js under a js directory.

spec/support/capybara.rb
runephilosof-abtion: There are these levels: Verbose, Info, Warning, or Error.

I assume you wouldn't want it to block on verbose and info messages.

Then do this:
```
          expect(error.level).not_to eq("SEVERE"), error.message
          expect(error.level).not_to eq("WARNING"), error.message
```

Then it is easy to remove the warning line, if it starts acting up (it will, if you add Stripe Checkout).

runephilosof-abtion (changes requested):

spec/support/capybara.rb
jeppester: FYI we did try with `console.log` and `console.info` to check if we could get rid of catch accidental console.log calls.

But the returned errors unfortunately did not seem to include `log` and `info` levels

jeppester:
> In `config/webpacker.yml` you can specify `source_path: app/frontend` to change the default name if you find it odd to nest non-js under a js directory.

We did consider this. We decided to keep the default path in an attempt to not open the door for a discussion about where to put things.

runephilosof-abtion (approved):

PR #90, opened 2020-08-31
Feature: Translate Devise views

bohme:
Asana-task: [Fraxx: Translate Devise views](https://app.asana.com/0/1142794766483633/1189212176701993/f)

Adds `devise-i18n` gem and locales for Danish and English. We also regenerated the views with:
`rails generate devise:i18n:views` to effectively extract all strings from the views.

We ran into some issues when writing view-specs to check if the translation worked. Devise expects to be able to use some inter helpers which aren't available when just rendering a view. 
That's what the:
```Ruby
      without_partial_double_verification do
        allow(view).to receive(:resource) { User.new }
        allow(view).to receive(:resource_name) { :user }
        allow(view).to receive(:devise_mapping) { Devise.mappings[:user] }
      end
```
Are there for 👍


app/helpers/application_helper.rb
```diff
@@ -1,4 +1,15 @@
...
+    :user
+  end
+
+  def resource
+    @resource ||= User.new
+  end
+
+  def devise_mapping
+    @devise_mapping ||= Devise.mappings[:user]
+  end
```
bohme, line 14: rubocop also doesn't like it 🤔

spec/rails_helper.rb
```diff
@@ -67,4 +67,5 @@
   # Include integration helpers for feature tests
   config.include Devise::Test::IntegrationHelpers, type: :feature
   config.include Devise::Test::ControllerHelpers, type: :controller
+  config.include Devise::Test::ControllerHelpers, type: :view
```
bohme, line 70: this might not be needed 🤔

config/locales/devise.da.yml
```diff
@@ -0,0 +1,146 @@
+# Additional translations at https://github.com/tigrish/devise-i18n/tree/master/rails/locales
```
jeppester, line 1: I think we can remove this file

config/locales/devise.en.yml
```diff
@@ -1,65 +1,148 @@
-# Additional translations at https://github.com/plataformatec/devise/wiki/I18n
+# Additional translations at https://github.com/tigrish/devise-i18n/tree/master/rails/locales
```
jeppester, line 1: And this file

config/locales/devise.da.yml
bohme: we can try 👍

bohme: you were right :D 💪

jeppester (approved):

spec/views/sessions/new.html.erb_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  context "with english locale" do
+    before { I18n.locale = :en }
+    after { I18n.locale = I18n.default_locale }
+
+    it "displays english text" do
+      stub_devise_helpers
+
+      render
+
+      expect(rendered).to include("Log in")
```
prrrnd, line 36: Wouldn't this test still pass with the locale set to `:da`, since the string `Log ind` include `Log in`?

bohme: 🤔

bohme: aww man... you're right 🌊 
thanks for sniping that one 💪

bohme: This should fix it:
```Ruby
expect(rendered).to have_selector("h2", exact_text: "Log in")
```

PR #93, opened 2020-08-31
Feature: Add parenthesis rubocop

bohme:
Asana-task: [chore: Add Rubocop cop for enforcing consistent parentheses ](https://app.asana.com/0/1142794766483633/1168070570718617/f)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -63,6 +63,30 @@ Style/FrozenStringLiteralComment:
 Style/Documentation:
   Enabled: false
 
+Style/MethodCallWithArgsParentheses:
+  IgnoredMethods: [
+    "to",
+    "to_not",
+    "require",
+    "require_relative",
+    "describe",
+    "it",
+    "context",
+    "exit",
+    "validate_presence_of",
+    "belong_to",
+    "have_many",
+    "raise",
+    "puts",
+    "desc",
+    "redirect_to",
+    "render"
+  ]
+  Enabled: true
+  Exclude:
+    - "db/migrate/*.rb"
+    - "Gemfile"
+
 Style/StringLiterals:
   EnforcedStyle: double_quotes
 
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -11,9 +11,9 @@ module Api
         user = User.find_by(id: request.headers["X-User-Id"])
 
         if user&.valid_token?(request.headers["Authorization"])
-          sign_in user
+          sign_in(user)
         else
-          head :unauthorized
+          head(:unauthorized)
         end
       end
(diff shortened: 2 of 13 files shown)


.rubocop.yml
```diff
@@ -63,6 +63,30 @@ Style/FrozenStringLiteralComment:
...
+    "context",
+    "exit",
+    "validate_presence_of",
+    "belong_to",
+    "have_many",
+    "raise",
+    "puts",
+    "desc",
+    "redirect_to",
+    "render"
```
jeppester, line 83: We are not making it easy for the developer to understand the rules with all these exceptions.

It makes me think if we really want to enforce this rule or not?

hcarreras: I'd be OK with no exceptions, and I also think that these exceptions are more readable without parenthesis and widely used.

I am against no parenthesis as a default, as this decreases readability.

hcarreras (approved):

MarkusAbtion:
> I am against no parenthesis as a default, as this decreases readability.

I think it helps readability by encasing parameters. 
It's quicker to scan methods

.rubocop.yml
bohme:
```Ruby
expect(that_thing).to(be_readable)
```
is a little weird... the defaults weren't choosen at random but as Hari points out are the ones that are more reable without parenthesis ^__^

jeppester:
I wish you could allow omitting the parentheses of outer calls, but disallow it for inner calls :thinking:

It would lead to a lot of less confusion, but keep the sleek syntax for most things.

.rubocop.yml
jeppester: From the list of exceptions, and the comments above it seems pretty clear to me that omitting parentheses is mostly readable when you've gotten used to it.

Can we expect everyone to have gotten used to this specific set of exceptions?
How would we continuously curate the list to make it up to date with what people have gotten used to?
Is it worth our time?

I remember struggling a lot with understanding the logic of rspec's DSL the first time I saw it.

bohme: The idea behind this is to ensure a uniform use of paranthesis, not to please people.
It is a rubocop rule, you type however you wanna type, you run rubocop and it turns it a uniformly formatted code.

I honestly don't see the problem, people get confused by code when they are new to the language, I'd find uniformly used parenthesis less confusing then random ones 🤷‍♀️

bohme: Think I'm willing to second Hari in that I'd be willing to remove all exceptions if need be 🤔

jeppester: It seems to me that the list of exceptions is there to please people.

But since I appear to be the only one who finds this long list of exceptions a bit odd, lets get this merged.

PR #94, opened 2020-08-31
Feature: Don't run seeds in release phase

bohme:
Asana-task: [Fraxx: Don't seed in the release phase action](https://app.asana.com/0/1142794766483633/1188627776772507/f)

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
 web: bundle exec puma -C config/puma.rb
-release: bundle exec rails db:migrate db:seed
+release: bundle exec rails db:migrate
--- a/app.json
+++ b/app.json
@@ -28,7 +28,7 @@
   "environments": {
     "review": {
       "scripts": {
-        "postdeploy": "bin/rake db:schema:load db:seed"
+        "postdeploy": "bundle exec rails db:seed"
       }
     }
   }
--- a/app.json
+++ b/app.json
@@ -24,5 +24,12 @@
   "formation": {},
   "name": "project-name-param",
   "scripts": {},
-  "stack": "heroku-18"
+  "stack": "heroku-18",
+  "environments": {
+    "review": {
+      "scripts": {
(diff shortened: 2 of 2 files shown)


prrrnd (approved):

prrrnd:
(shouldn't there be a worker as well for sidekiq?)

bohme:
(first after we have added sidekiq to this project)

jeppester:
Won't this change cause the review apps to have their DBs seeded?

If so I think we should also add a review app script that does the seeding.

hcarreras (approved):

jeppester (changes requested):
As stated in my comment above we should still seed review apps

bohme:
I agree, that's why I added this commit: https://github.com/abtion/muffi/pull/94/commits/1a6c553d023e980840d3407d5946c3a6915e9c4e

jeppester:
Oh, I didn't notice that. I'll recreate the review app and see if it works as intended, then lets merge if it does :-)

jeppester:
It didn't work as intended :disappointed:
The postdeploy script did not appear to run at all

bohme:
yea looks like stuff blew up... I suspect it to be the `schema:load` part 🤷‍♀️

jeppester:
It appears to be working now :+1:, I'll rebuild the review app once more to do one final check.
It worries me that I can't seem to find a log of the postdeploy script in heroku's UI. Maybe I'm not good at looking for it? :thinking:

jeppester:
The seed user magically appears shortly after the release script is finished running. So it works :-)

bohme:
does this work?
https://help.heroku.com/LXRRBMM3/how-do-i-view-postdeploy-script-output#:~:text=This%20output%20is%20available%20via,View%20initial%20app%20setup%22%20option.

jeppester (approved):

PR #96, opened 2020-09-02
app.json: Don't require HEROKU_APP_NAME env var

jeppester:
Slightly related to this:
https://app.asana.com/0/1142794766483633/1190999859256033

We only use HEROKU_APP_NAME in review apps, and it's set automatically for those.


bohme (approved):
nice 🌊

PR #98, opened 2020-10-26
Meta: Add action for auto-merging dependabot

MarkusAbtion:
We're using the [Dependabot Auto Merge](https://github.com/marketplace/actions/dependabot-auto-merge) action
from the Github Marketplace.
Also, we've added the `DEPENDABOT_TOKEN` as a secret for authorization.

We cannot trigger this action manually, so we're waiting for a security update to happen
to test it.

Associated task: https://app.asana.com/0/1142794766483633/1187208358171570/f

Co-authored-by: Mathias Jean Johansen <mj@abtion.com>

### DIFF
--- a/.github/auto-merge-dependabot.yml
+++ b/.github/auto-merge-dependabot.yml
@@ -1,0 +1,14 @@
+name: auto-merge
+
+on:
+  pull_request:
+
+jobs:
+  auto-merge:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


jeppester:
Will the dependabot token need to be updated on a per-project basis?
If so we might want to document it in the README

mathias-abtion:
@jeppester: Good point! This most likely needs to be configured per project, so we will mention it in the README in a commit later today.

PR #99, opened 2020-10-26
Use `schema.rb` instead of `structure.sql`

mathias-abtion:
It seems that we are rarely using features that require a `structure.sql`, and we've dropped the `structure.sql` file in many of our newer projects so this pull request converts Muffi from using `structure.sql` to `schema.rb`.

## Asana link
[Use schema.rb instead of structure.sql?](https://app.asana.com/0/1142794766483633/1190986905068737/f)


jeppester (approved):
Fantastic :+1:

PR #102, opened 2020-10-26
Meta: Add details about Rubocop to README

MarkusAbtion:
Adds missing details about Rubocop to README.example.

## Asana task
There is no Asana task for this.


jeppester (approved):
:+1:

PR #103, opened 2020-10-27
Design: Update login page

MarkusAbtion:
This PR introduces the column design as [shown on Figma](https://www.figma.com/file/JwmoccxabsSdOkPeYDI1I5/Muffi-Template?node-id=4866%3A23369
)

It also adds Danish locale files for devise and general translations.

Based on a [discussion on Slack](https://abtion.slack.com/archives/CPW7LUT8B/p1603794093002900), we made these changes to the login page instead of, as the Asana task requires, the homepage.

## Asana task:
[update default muffi homepage to our default design](https://app.asana.com/0/1142794766483633/1174657289308207)

## Screenshot:
Screen Shot 2020-10-27 at 16 12 54


jeppester (changes requested):
Overall I think you've done a really good job here :+1: 

I found a few things the deviates from the figma design:

- The smallest screen size doesn't appear to be implemented:
image
- The input field corners are all rounded in the design, but not in the implementation
- The font doesn't appear to be correct
- The "forgot password" link has a different color

I'm not sure that all of these things should look exactly like in the design. But I think you should get in touch with Jacob and maybe add some final touches together.

jeppester:
I've rebased the PR and improved few things:
- The right font has been added
- The spacing between elements are now slightly more like the design
- The input border radiuses are now consistent with the design
- The outer margins have been adjusted to look better in phones (not quite like the design, but as close a we can get with simple bootstrap classes)

More work can be done here, but I think it's best to get this merged as it is and when improve upon it later.

PR #104, opened 2020-11-27
Feature: Add script for renaming files

bohme:
This PR adds a little Ruby script named `bin/rename-files` which gets called in `bin/setup`.

It will assume the name of the current directory to be the project name and do the search/replace for the differently cased project names mentioned in the `README` on every file known to git.

In a quick'n'dirty fashion I monkey-patched `String` with some case logic from `Rails`. 👍 

Not sure where to test this 🤔 

```sh

$ time bin/rename-files

real    0m0.281s
user    0m0.120s
sys     0m0.129s
```

Associated task:
https://app.asana.com/0/1142794766483633/1199360272751221/f


bin/rename-files
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env ruby
+require 'fileutils'
+
+class String
+  def param_case
+    gsub(/::/, "/")
+      .gsub(/([A-Z]+)([A-Z][a-z])/, '\1_\2')
+      .gsub(/([a-z\d])([A-Z])/, '\1_\2')
```
jeppester, line 8: I might be wrong, but it looks to me like this will do snake case rather than param case.

I would have expected the replacement to be `'\1-\2'` rather than '\1_\2'.

prrrnd:
Was it such a pain to search and replace in the project that we needed a script for it? Seems like we'll need to start quite a few projects to make up for the time invested in this.

jeppester:
I'm finalizing this PR and merging it to get it out of the way.

bin/setup
```diff
@@ -13,6 +13,9 @@ FileUtils.chdir APP_ROOT do
   # This script is idempotent, so that you can run it at anytime and get an expectable outcome.
   # Add necessary setup steps to this file.
 
+  puts '== Checking project file names =='
+  system! 'bin/rename-files'
+
```
jeppester, line 18: `bin/setup` is for setting up the project itself, not for setting up the template.
If we want a script for setting up the template itself I think it belongs elsewhere.

jeppester:
I've fixed some issues with this and merged it.
We can implement feedback later if necessary.

PR #110, opened 2020-12-23
Simplify Capybara setup

prrrnd:
Keep only the headless chrome driver, as it is the one most developers
will use.

Move the javascript related support helper to its own file.

https://app.asana.com/0/1142794766483633/1199195024410608


jeppester:
:1st_place_medal: :trophy:

PR #111, opened 2020-12-23
Add stylelint for SCSS linting

prrrnd:
https://app.asana.com/0/1142794766483633/1193868397570278

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,0 +1,20 @@
+{
+  "extends": "stylelint-config-sass-guidelines",
+  "rules": {
+    "selector-max-id": 1,
+    "max-nesting-depth": 3,
+    "selector-max-compound-selectors": 4,
+    "selector-no-qualifying-type": null,
+    "string-quotes": "double",
+    "order/properties-alphabetical-order": null,
+    "declaration-bang-space-before": null,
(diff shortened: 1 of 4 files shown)


.stylelintrc.json
```diff
@@ -0,0 +1,20 @@
...
+    "string-quotes": "double",
+    "order/properties-alphabetical-order": null,
+    "declaration-bang-space-before": null,
+    "declaration-colon-space-after": null,
+    "no-missing-end-of-source-newline": null,
+    "number-leading-zero": null,
+    "selector-list-comma-newline-after": null,
+    "indentation": null,
+    "color-hex-case": null,
+    "function-comma-space-after": null
```
jeppester, line 17: I'm not sure I understand why we need all these custom rules, but I guess there are some thoughts behind them.

runephilosof-abtion: They were turned off to avoid having too many changes in the same commit.
They should be turned on again (remove the line setting it to null).
But this was never done in the project it was extracted from...
We need to iterate on it in another project.

jeppester: I think we should have some degree of freedom from this rule of thumb, for the purpose of setting a higher standard for future projects.

PR #115, opened 2020-12-23
Set up rack-attack

prrrnd:
Allow blocking by IP address
Allow limiting the number of requests


jeppester:
Have we made any attempts to get blocked, to ensure that it works as intended?

prrrnd:
We have, however we were not able to set up a test for that scenario. We suspect that we need to set up some caching for it to work, based on the rack-attack documentation. We'll get back at it later.

jeppester:
I'll rebase this and add tests

PR #116, opened 2020-12-23
Ruby 2.7.1

prrrnd:

https://app.asana.com/0/1142794766483633/1199666322538309


DeWinther (approved):
Looks good! 💪

runephilosof-abtion:
This adds a ton of warnings when running rspec.
If they aren't easy to fix, then I would prefer to revert back to ruby 2.6, and create task for fixing the warnings before upgrading.

prrrnd:
All warnings are coming from external dependencies (mostly capybara and devise). I don't think we should stick to 2.6, but update the dependencies instead. Good catch though 👍🏼

runephilosof-abtion:
My point is that I don't think we can remove all warnings by upgrading dependencies, because the the gems aren't ready for Ruby 3.0.
Ruby 2.7 emits warnings about stuff that is planned to be removed in Ruby 3.0 ([Separation of positional and keyword arguments](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/)).

Ruby 2.6.x is still supported. So I think we should stick to that version until the dependencies have been fixed and don't emit warnings.
The problem with switching too early is that people won't notice, when they add code that emits a warning, because it drowns in the existing warnings.

prrrnd:
@runephilosof-abtion https://github.com/abtion/muffi/pull/118 fixes all warnings when running rspec.
2.7 is over a year old now, if the dependencies have not been updated after a year, perhaps they should be re-evaluated.

jeppester:
I hope we'll get [dependabot working](https://app.asana.com/0/1142794766483633/1187208358171570) soon. It could have prevented this whole talk.

I also think we should consider failing tests that write to stderr or stdout. I've added a discussion card for that:
https://app.asana.com/0/1142794766483633/1199524350907707

runephilosof-abtion:
@prrrnd What gem do you suggest to replace [rspec-mocks](https://github.com/rspec/rspec-mocks/issues/1326) with?
@hcarreras I think we should re-evaluate whether Abtion should invest a couple of pair days in fixing the rspec-mocks gem.

hcarreras:
At first glance, it seems that the `rspec-mocks` gem would not be that easy to fix.
This is definitely a project that can be taken in the Friday initiatives that we are trying to push forward 💪
Ideally, developers would contribute to open source as part of their Friday activities. Abtion will spend time organizing those activities. We can talk about it during our next 1:1 if you want to organize it.

jeppester:
Maybe this is a good opportunity for abtion to contribute to rspec

PR #120, opened 2021-02-22
Dependabot: Update all versions, not only security

runephilosof-abtion:
By default Dependabot only creates pull requests for security updates.
With this configuration it will create pull requests for all updates daily.
The npm package caniuse-lite is updated daily (upstream), so to limit
the notifications, npm is configured to a weekly schedule.


bohme (approved):
Very nice 🤖

PR #135, opened 2021-02-22
Chore: rails app update

bohme:
Runnign `rails app:update` and bringing the app up to `6.1` speed.


config/application.rb
```diff
@@ -14,51 +14,24 @@
...
 # require "rails/test_unit/railtie"
 
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
+
 module ProjectNamePascal
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults(6.0)
```
runephilosof-abtion, line 27: It seems we forgot to change the defaults, even though we were aware of it.

config/initializers/content_security_policy.rb
```diff
@@ -1,33 +1,32 @@
...
-  policy.frame_ancestors(:none)
-  policy.img_src(:self, :https, :data)
-  policy.object_src(:none)
-  policy.script_src(:self)
-  policy.style_src(:self)
-  # If you are using webpack-dev-server then specify webpack-dev-server host
-  if Rails.env.development?
-    policy.connect_src(:self, :https, "http://localhost:3035", "ws://localhost:3035")
-  end
+# Rails.application.config.content_security_policy do |policy|
```
runephilosof-abtion, line 8: We accidentally turned off CSP :face_holding_back_tears:

PR #136, opened 2021-02-22
Chore: Update rubocop slightly

bohme:
Rubocop is awfully out of date, this brings it slightly closer to the modern day.

Fixes the following warning:
```
Warning: RSpec/DescribeClass does not support IgnoredMetadata parameter.

Supported parameters are:

  - Enabled
```


runephilosof-abtion (approved):
Loving it.

PR #174, opened 2021-03-15
Attempt to fix failing dependabot PRs

jeppester:
Archive coverage for code-climate action

### DIFF
--- a/.github/workflows/auto-merge-dependabot.yml
+++ b/.github/workflows/auto-merge-dependabot.yml
@@ -1,7 +1,7 @@
 name: auto-merge
 
 on:
-  pull_request:
+  pull_request_target:
 
 jobs:
   auto-merge:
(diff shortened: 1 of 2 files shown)


runephilosof-abtion:
I don't quite understand why this stopped working, or rather the timing of it.
It was working March 8th.
The only change mentioned on the github.blog that has anything to do with this is https://github.blog/2021-03-08-github-security-update-a-bug-related-to-handling-of-authenticated-sessions/ which requires users to log in again.
But it does not mention invalidating the tokens.

No mention of changing access to secrets.
The change with pull_request_target was supposed to be in effect from March 1st https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/, so it is a bit weird that auto merge worked on March 8th.

PR #285, opened 2021-05-18
Use rails server and port 3000 by default for dev

jeppester:
I've changed the port to 3000 because that is most comment (as far as I know), and more importantly to make it match the port we've set for mailer urls in our development configuration.

```rb
  config.action_mailer.default_url_options = { host: "localhost", port: 3000 }
```

Asana: https://app.asana.com/0/1142794766483633/1200257919012205

### DIFF
--- a/README.example.md
+++ b/README.example.md
@@ -119,7 +119,7 @@ The hooks are symlinked meaning all the changes to the repo hooks will automatic
 
 ## Day-to-day
 
-- Run the server: `heroku local` and [http://localhost:5000](http://localhost:5000)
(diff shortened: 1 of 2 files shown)


runephilosof-abtion (approved):
I aggree

prrrnd:
Shouldn't we have updated the mailer port too with `ENV.fetch("PORT", 3000)`?
Now if we set a custom port in `.env.local`, the mailer stays at 3000.

runephilosof-abtion:
> Shouldn't we have updated the mailer port too with `ENV.fetch("PORT", 3000)`?
> Now if we set a custom port in `.env.local`, the mailer stays at 3000.

Sound reasonable

PR #351, opened 2021-06-22
Enable more relevant eslinters.

runephilosof-abtion:
Recommended by the testing library author https://kentcdodds.com/blog/common-mistakes-with-react-testing-library


hcarreras (commented):
Looks good

PR #352, opened 2021-06-22
Configure eslint to allowEscape.

runephilosof-abtion:
Prettier rewrites it to single quotes, when the strings contain double quotes.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,7 +12,7 @@ module.exports = {
     // enable additional rules
     "indent": ["error", 2],
     "linebreak-style": ["error", "unix"],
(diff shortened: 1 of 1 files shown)


runephilosof-abtion:
Should probably use 
eslint:
extends: ["eslint:recommended", "prettier"],
package.json:
"eslint-config-prettier"

instead

hcarreras (approved):

jeppester:
I have this task in the icebox that's about the same thing:
https://app.asana.com/0/1142794766483633/1200306308003616

I think `eslint-config-prettier` is a better solution, since it eases the maintenance burden for us, and probably solves issues that we haven't hit yet.

PR #392, opened 2021-07-14
 CI: Archive coverage reports on failures

runephilosof-abtion:
Since we require 100% covered, there is no reason to use time on archiving the reports unless it failed.


jeppester (approved):
Great idea 👍

jeppester:
I merged this PR too fast.

Coverage is uploaded as an artifact so that the code climate workflow can retrieve it and upload it.
I've reverted the commit in the main branch to fix code climate (I didn't know there's a "revert" button in here).

PR #393, opened 2021-07-14
Readme: New project instructions

runephilosof-abtion:
Maybe we should choose one of the methods. I prefer the second one, retaining the history.


jeppester:
Sometimes we add projects as remote to be able to cherry-pick commits into muffi.

I wonder if we can be 100% sure that objects from those other projects (apart from the chosen commits) won't get into a new project if we retain the history.

Apart from that worry I'm 100% for cloning the git repo on new projects.

PR #429, opened 2021-08-09
Also archive coverage reports on failure

runephilosof-abtion:
If the failure is caused by too low coverage, then the report is important to have.


runephilosof-abtion:
Follow up for #392

jeppester (approved):
Good point :+1:

PR #440, opened 2021-08-12
Add bullet gem

jeppester:
I would like to have had tests for this, but we don't currently have any related tables in the our schema.
So to be able to test, I would have had to make some sort of pro-forma relationship in our schema.

I decided to go with having no tests, since I believe that to be the "lesser evil".
If you have better ideas, let me know.

### DIFF
--- a/config/initializers/bullet.rb
+++ b/config/initializers/bullet.rb
@@ -16,6 +16,7 @@ Rails.application.configure do
 
     if Rails.env.test?
       Bullet.raise = true # raise an error if n+1 query occurs
+      Bullet.unused_eager_loading_enable = false
     end
   end
 end
--- a/config/initializers/bullet.rb
+++ b/config/initializers/bullet.rb
@@ -4,18 +4,18 @@ return if Rails.env.production?
 
 Rails.application.configure do
   config.after_initialize do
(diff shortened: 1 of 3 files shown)


config/initializers/bullet.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    if Rails.env.development?
+      Bullet.alert = true
+      Bullet.console = true
+      Bullet.rails_logger = true
+      Bullet.add_footer = true
+    end
+
+    if Rails.env.test?
+      Bullet.raise = true # raise an error if n+1 query occurs
```
runephilosof-abtion, line 18:
```suggestion
      Bullet.raise = true # raise an error if n+1 query occurs
      Bullet.unused_eager_loading_enable = false
```

Because you often make tests that does not contain enough data to merit the `includes` that are needed with a full dataset.

runephilosof-abtion (changes requested):
Awesome.

About the tests. Why test whether a gem works, those tests belong in the gem itself.

jeppester:
@runephilosof-abtion Better?

runephilosof-abtion (approved):

PR #444, opened 2021-08-12
README: Change CI configuration so that we use a flow that automatically creates addons

jeppester:
I investigated this issue:
https://app.asana.com/0/1142794766483633/1200529919197770

As apparent from the comments on the card it turned out to be a bit of a rabbit hole.

TLDR: Heroku doesn't create addons when you first create an app directly from the pipeline and then deploy it through github. We need to use https://heroku.com/new and tell it to use muffi as template.

I've updated the README to facilitate that workflow, so that hopefully it will be less of a bumpy ride when we create muffi-based apps going forward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,7 @@ Proprietary and confidential.
 2. Create the production app the same way as the staging app, only this time:
    - Name the app `<PROJECT-NAME>-production`.
    - Add the app to the `<PROJECT-NAME>` pipeline rather than creating one.
+   - Add the app to the "production" stage of the pipeline
 3. When app creation has succeeded, press "manage app" to get to the app, then select the pipeline in the breadcrumb
 4. Connect the pipeline to the project's github repo. (There should be a "Connect to Github" button).
 5. Enable review apps.
--- a/README.md
+++ b/README.md
@@ -116,7 +116,8 @@ Proprietary and confidential.
    5. Fill in required env vars (read the field descriptions for guidance).
    6. Press the deploy button.
    7. Take a 5 minute break while heroku sets up app and pipeline.
-   8. Sometimes the app will build, but it doesn't go on with the next steps and there is a not very informative error message. If so the app has likely been deployed, but not added to the pipeline. You will then have to manually create the pipeline and add the app to it.
+   8. Sometimes the app will build, but it doesn't go on with the next steps and there is a not very informative error message.
+      - If so the app has likely been deployed, but not added to the pipeline. You will then have to manually create the pipeline and add the app to it.
 2. Create the production app the same way as the staging app, only this time:
    - Name the app `<PROJECT-NAME>-production`.
    - Add the app to the `<PROJECT-NAME>` pipeline rather than creating one.
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ Proprietary and confidential.
 2. Create the production app the same way as the staging app, only this time:
    - Name the app `<PROJECT-NAME>-production`.
    - Add the app the `<PROJECT-NAME>` pipeline rather than creating one.
-3. Go to the pipeline "manage app" -> select the pipeline in the breadcrump.
+3. Go to the pipeline "manage app" -> select the pipeline in the breadcrumb.
 4. Connect the pipeline to the project's github repo. (There should be a "Connect to Github" button).
 5. Enable review apps.
 6. Add config vars for review apps:
--- a/README.md
+++ b/README.md
@@ -108,16 +108,27 @@ Proprietary and confidential.
 
 ### Configure CD
 
-1. Log in to the [Heroku dashboard](https://dashboard.heroku.com)
-2. Create a Pipeline, and two new applications
-   - Staging: `<PROJECT-NAME>-staging`
-   - Production: `<PROJECT-NAME>-production`
-3. Turn on "Review Apps" from the Pipeline's page
-4. Set Heroku config vars:
-   - `MAIL_FROM` (review, staging and production).
-   - `DOMAIN_NAME` (ONLY staging and production)
-   - `SEED_ADMIN_EMAIL` (review)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -108,16 +108,27 @@ Proprietary and confidential.
...
+   6. Press the deploy button.
+   7. Take a 5 minute break while heroku sets up app and pipeline.
+   8. Sometimes the app will build, but it doesn't go on with the next steps and there is a not very informative error message. If so the app has likely been deployed, but not added to the pipeline. You will then have to manually create the pipeline and add the app to it.
+2. Create the production app the same way as the staging app, only this time:
+   - Name the app `<PROJECT-NAME>-production`.
+   - Add the app the `<PROJECT-NAME>` pipeline rather than creating one.
+3. Go to the pipeline "manage app" -> select the pipeline in the breadcrump.
+4. Connect the pipeline to the project's github repo. (There should be a "Connect to Github" button).
+5. Enable review apps.
+6. Add config vars for review apps:
```
prrrnd, line 126: Can we automate this process with `heroku config:set`, or generate some from the app.json?

README.md
```diff
@@ -108,16 +108,27 @@ Proprietary and confidential.
...
+   3. Select europe as region.
+   4. Create a new pipeline for the project `<PROJECT-NAME>`.
+   5. Fill in required env vars (read the field descriptions for guidance).
+   6. Press the deploy button.
+   7. Take a 5 minute break while heroku sets up app and pipeline.
+   8. Sometimes the app will build, but it doesn't go on with the next steps and there is a not very informative error message. If so the app has likely been deployed, but not added to the pipeline. You will then have to manually create the pipeline and add the app to it.
+2. Create the production app the same way as the staging app, only this time:
+   - Name the app `<PROJECT-NAME>-production`.
+   - Add the app the `<PROJECT-NAME>` pipeline rather than creating one.
+3. Go to the pipeline "manage app" -> select the pipeline in the breadcrump.
```
MarkusAbtion, line 123: I'm not sure what this line means. 
Is there a "Manage app"-button in the pipeline view?

README.md
```diff
@@ -108,16 +108,27 @@ Proprietary and confidential.
...
-   - `SEED_ADMIN_INITIAL_PASSWORD` (review)
+1. Create staging heroku app based on the muffi template.
+   1. Visit: https://dashboard.heroku.com/new?org=abtion&template=https%3A%2F%2Fgithub.com%2Fabtion%2Fmuffi
+   2. Fill in the name with `<PROJECT-NAME>-staging`.
+   3. Select europe as region.
+   4. Create a new pipeline for the project `<PROJECT-NAME>`.
+   5. Fill in required env vars (read the field descriptions for guidance).
+   6. Press the deploy button.
+   7. Take a 5 minute break while heroku sets up app and pipeline.
+   8. Sometimes the app will build, but it doesn't go on with the next steps and there is a not very informative error message. If so the app has likely been deployed, but not added to the pipeline. You will then have to manually create the pipeline and add the app to it.
```
bohme, line 119:
```suggestion
   8. Sometimes the app will build, but it doesn't go on with the next steps and there is a not very informative error message.
      - If so the app has likely been deployed, but not added to the pipeline. You will then have to manually create the pipeline and add the app to it.
```

bohme: Very long lines in lists are hard to read 👍 
(not 100% sure if I got the spacing right)

bohme (commented):
Could we automate all of this with a bash script utilising heroku-cli?

README.md
jeppester: I thought about it, but I'm not sure we can do it in a good way.

I might be bad at looking, but I can't find a way to update preview app env vars through the heroku command line.

What we can do is up app.json to generate passwords for review apps, but then a new password will be generated for each review app.

bohme (commented):
Could we also ensure that all of this is accessible for non-admins?

README.md
bohme: this one maybe?
https://devcenter.heroku.com/articles/heroku-cli-commands#heroku-ci-config-set
Unsure if CI config vars are the same as just pipeline config vars 🤷‍♀️

jeppester: Based on https://github.com/abtion/muffi/pull/444#pullrequestreview-728521107 it appears that I was in fact not looking the right place.

As long as we can ensure that every project (not review app) has a unique password, then I guess it's fine.

jeppester:
> Could we automate all of this with a bash script utilising heroku-cli?
> https://devcenter.heroku.com/articles/heroku-cli-commands#heroku-ci-config-set

We might.

I've tried using `heroku pipelines:setup` which should do a lot of what we want, but I currently does nothing.
Might be just today, or something that happens often :man_shrugging: .

What I've done today is to actually go through the whole flow and find things that work, but maybe things work differently tomorrow...

I'll revisit this later and see if things have changed. I still haven't found a way to set review app env vars through the CLI though. - only CI env vars.

README.md
jeppester: The "Manage app" button shows up when then creation has completed, My hope was to guide the user from the creation page to the next steps in the guide. I can see that I could have done a better job :-).

How about?:
```
3. When app creation has succeeded, press "manage app" to get to the app, then select the pipeline in the breadcrumb
```

MarkusAbtion: That works for me

bohme:
I would like to volunteer to take over this task ^___^

The only thing I love more than muffi is the heroku-cli (and purple... and vim) so being able to work on this tasks would make me the happiest camper what ever walked the earth 💜

hcarreras (approved):
Looks good. Let merge this and see if Thor comes up with an improvement PR

jeppester:
I've gone through the steps again tonight and added an additional bullet, to ensure that the production app is added to the correct pipeline stage.

I've also tried the `heroku pipelines:setup` command again, which I believe to be the only command that uses and `app.json` file. I still can't get it to do anything, though it's output looks correct.

I'm merging this for now.

PR #454, opened 2021-08-18
Update bootstrap v4 -> v5

jeppester:


runephilosof-abtion (commented):
Nice.
This one should have been removed as well
https://github.com/abtion/muffi/blob/main/frontend/stylesheets/globals/_accessibility.scss
.visually-hidden is part of bootstrap now

runephilosof-abtion (commented):
Also update the navbar (data-bs-toggle, data-bs-target, ml-* → ms-*)

PR #484, opened 2021-08-30
Fix testing rack attack.

runephilosof-abtion:
When the bucket is emptied the requests are allowed again.
Which could happen just before the last request.
So we want to test that some of the requests have been throttled instead of only the last one.

### DIFF
--- a/spec/rack/attack_spec.rb
+++ b/spec/rack/attack_spec.rb
@@ -70,7 +70,7 @@ RSpec.describe Rack::Attack, type: :request do
             statuses[i] = response.status
           end
 
-          expect(statuses).not_to(include(429))
+          expect(statuses).to_not include(429)
         end
       end
     end
--- a/spec/rack/attack_spec.rb
+++ b/spec/rack/attack_spec.rb
@@ -40,7 +40,7 @@ RSpec.describe Rack::Attack, type: :request do
   describe "IP address throttling" do
     let(:limit) { 20 }
 
-    context "when the number of requests is lower than the limit" do
+    context "when the number of requests does not exceed the limit" do
       it "does not change the request status" do
         limit.times do
           get "/", env: { REMOTE_ADDR: "1.2.3.4" }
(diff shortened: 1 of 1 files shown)


jeppester (approved):
Awesome! 🏆

While I'm not entirely happy that I didn't see this problem myself when I made the test, I'm really happy that you found a reasonable explanation for the random failures. It makes a lot of sense! 🙂

prrrnd (approved):
Looks good, and nice comments.
Looking at the file, this test above isn't very clear:

```ruby
    context "when the number of requests is lower than the limit" do
      it "does not change the request status" do
        limit.times do # <--- Wouldn't (limit - 1) be more explicit and match the context def?
          get "/", env: { REMOTE_ADDR: "1.2.3.4" }
          expect(response.status).to_not eq(429)
        end
      end
    end
```

jeppester:
> Looks good, and nice comments.
> Looking at the file, this test above isn't very clear:
> 
> ```ruby
>     context "when the number of requests is lower than the limit" do
>       it "does not change the request status" do
>         limit.times do # <--- Wouldn't (limit - 1) be more explicit and match the context def?
>           get "/", env: { REMOTE_ADDR: "1.2.3.4" }
>           expect(response.status).to_not eq(429)
>         end
>       end
>     end
> ```

I think the description should be changed to "... lower than or equal to the limit".

I don't like changing the number of requests since it makes it less clear when the limit is exceeded.

prrrnd:
But then it wouldn't actually test both scenarios (lower than and equal to the limit). We would need two tests 🙂

jeppester:
> But then it wouldn't actually test both scenarios (lower than and equal to the limit). We would need two tests slightly_smiling_face

There's an expectation for every request, so I wouldn't say that we don't test both scenarios.

runephilosof-abtion:
Isn't it just a matter of adjusting the text instead
`context "when the number of requests does not exceed the limit" do`
or 
`context "when the number of requests equals the limit" do`

jeppester:
I'm merging this, we can make changes later if we need it

PR #486, opened 2021-09-02
Update babel config js

runephilosof-abtion:
Make babel.config.js look more like Rails default

Remove require(...).default. I do not remember why it was once added. I believe it might have been added by some version of prettier or eslint.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -18,7 +18,7 @@ module.exports = function(api) {
   return {
     presets: [
       isTestEnv && [
-        require('@babel/preset-env').default,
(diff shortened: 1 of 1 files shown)


jeppester (approved):
> Make babel.config.js look more like Rails default
> 
> Remove require(...).default. I do not remember why it was once added. I believe it might have been added by some version of prettier or eslint.

It was part of the initial commit to muffi - before eslint was added. My guess is that it was how the default webpacker configuration looked back then.

Great effort keeping the configuration up-to-date :+1:

PR #495, opened 2021-09-08
Ignore RoutingError in Rollbar

DeWinther:
Routing errors should always result in a 404, which we are handling with
a 404 page (it could be prettier), so this is not actually something we
want to be reported in Rollbar

For the future, I would think we would need a proper 404 page we can
reuse across all of our applications instead of rails standard page

### DIFF
--- a/config/initializers/rollbar.rb
+++ b/config/initializers/rollbar.rb
@@ -33,6 +33,7 @@ Rollbar.configure do |config|
   #
   # You can also specify a callable, which will be called with the exception instance.
   # config.exception_level_filters.merge!('MyCriticalException' => lambda { |e| 'critical' })
+  config.exception_level_filters["RoutingError"] = "ignore"
 
   # Enable asynchronous reporting (uses girl_friday or Threading if girl_friday
   # is not installed)


hcarreras (approved):
Good first step and I agree with a nicer 404 page in Muffi 👍

bohme (approved):
I agree, nice find 👍 

🎉 https://app.asana.com/0/1142794766483633/1200942401806312/f

prrrnd:
I'm not sure I agree with this. Why are users landing on 404s in the first place? Is that a consequence of bad UX / navigation? Are we putting ourselves in a position where we can miss actual issues? I could imagine something along the lines of `link_to "xxxx", model_path(model)` where `model` is nil?

jeppester:
> I'm not sure I agree with this. Why are users landing on 404s in the first place? Is that a consequence of bad UX / navigation? Are we putting ourselves in a position where we can miss actual issues? I could imagine something along the lines of `link_to "xxxx", model_path(model)` where `model` is nil?

I think this is a very valid concern. I'm however unsure if this change ignores `RecordNotFound`-errors. I don't recall such errors showing up as `RoutingError`s in rollbar, but I might remember wrong. Can we check this @DeWinther ?

hcarreras:
@prrrnd I agree, it is a valid concern, however, it ends being too much noise. I do not want to get informed if someone decides to type something in their navigation bar. We should find a better way to get informed about the scenario that you describe.

@jeppester I believe that @DeWinther and I checked this already. `ActionController::RoutingError` is been rescued here https://github.com/rails/rails/blob/c2e42a26ecb08118bde6c0b8776bd255ed60f828/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb and it renders a view (based on the template).

We checked all the places where RoutingError was used in Rails.

jeppester:
If we are certain the `RecordNotFound`-errors are still reported, then I think it's fine to merge this

runephilosof-abtion:
Some work has been done on this before
@bohme Didn't you create some rule in Rollbar or Sentry to ignore 404 when referer is empty?

bohme:
I remember there was something we did similar to this... but I cannot for the life of me find it again >__<

jeppester:
@bohme Did you do a thorough search for the mentioned code?

DeWinther:
@bohme was this the one you did some work on? https://app.asana.com/0/1146601270380196/1160056597795120/f

This ticket was also where the idea came from @jeppester

PR #503, opened 2021-09-13
Policies: Replace CanCanCan with Pundit

MarkusAbtion:
- Use Pundit instead of Cancancan.
- Correct rubocop issues


jeppester (approved):
Great work :+1:

bohme (approved):
looks good 💪

PR #513, opened 2021-09-14
Localization: Update i18n-tasks config to only read locales in folders

MarkusAbtion:
This means it will not recognize `config/locales/da.yml` anymore.


jeppester:
> This means it will not recognize `config/locales/da.yml` anymore.

Are you sure?

I can't say for certain how the glob is resolved by i18n-tasks, but I would expect `folder/**/something` to look in the folder itself + subfolders.
billede

MarkusAbtion:
> > This means it will not recognize `config/locales/da.yml` anymore.
> 
> Are you sure?
> 
> I can't say for certain how the glob is resolved by i18n-tasks, but I would expect `folder/**/something` to look in the folder itself + subfolders.
> billede

I'm not sure, no. I might not know my asterisks 🙃

runephilosof-abtion:
> I can't say for certain how the glob is resolved by i18n-tasks, but I would expect `folder/**/something` to look in the folder itself + subfolders.

You are correct.
I checked i18n-tasks today.

PR #516, opened 2021-09-16
React FE setup

jeppester:


runephilosof-abtion:
Couldn't this have been merged without the tailwind change?
It feels like a lot to have end up in a single squashed commit.

jeppester:
> Couldn't this have been merged without the tailwind change?
> It feels like a lot to have end up in a single squashed commit.

We thought about doing that, but we don't want to merge until we've tested webpacker 6 more thoroughly.
We might have to go back and fix issues.

For now this PR is just meant for ensuring that CI passes.

PR #532, opened 2021-09-20
Run `rails assets:precompile` instead of `webpack`.

runephilosof-abtion:
Because in some cases webpack will succeed where precompile will fail.
It is slower, but not slower than rspec, and now it is run in parallel.

https://app.asana.com/0/1142794766483633/1201016019690455

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -161,6 +161,39 @@ jobs:
           bundle exec rake db:migrate
       - name: check git diff
         run: git diff --exit-code db/schema.rb
+
+  assets-precompile:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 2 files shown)


runephilosof-abtion:
It is failing because of the recent sass gem upgrade.
Which is the motivation for this change to the CI, it should fail, because Heroku will fail.

jeppester (approved):
@runephilosof-abtion Great idea. If you have the time would you mind reverting the sass upgrade in this PR and then merging?

jeppester:
I reverted the sass commit. Since the CI now succeeds I'm merging this.

runephilosof-abtion:
Thanks for helping this to the finish line

PR #533, opened 2021-09-21
React fe setup

jeppester:


app/views/layouts/application.html.erb
```diff
@@ -6,13 +6,14 @@
...
-    <%= javascript_pack_tag "application" %>
     <%= stylesheet_pack_tag "application" %>
   </head>
 
   <body class="<%= "#{controller.controller_name} #{@body_class}".strip %> h-full">
     <main class="w-full h-full">
       <%= yield %>
     </main>
+
+    <%= javascript_pack_tag "application" %>
```
jeppester, line 18: Load async in top

yarn.lock
```diff
@@ -1248,6 +1248,11 @@
     minimatch "^3.0.4"
     strip-json-comments "^3.1.1"
 
+"@heroicons/react@^1.0.4":
```
jeppester, line 1251: This is not used

app/views/layouts/application.html.erb
jeppester: Load "defer" in top was the answer, and also what webpacker does by default :champagne:

PR #638, opened 2021-11-25
Specs/change to system tests

reinisla:
Asana: https://app.asana.com/0/1142794766483633/1200773528997220/f

-----------------

Switch to system tests introduces new deprecation errors:

https://app.asana.com/0/0/1201426372899169/f

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -67,8 +67,4 @@ RSpec.configure do |config|
   # Include integration helpers for feature tests
   config.include(Devise::Test::IntegrationHelpers, type: :system)
   config.include(Devise::Test::ControllerHelpers, type: :controller)
-
-  config.before(:each, type: :system) do
-    driven_by(:selenium_chrome_headless)
-  end
 end
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -65,6 +65,10 @@ RSpec.configure do |config|
   # config.filter_gems_from_backtrace("gem name")
 
   # Include integration helpers for feature tests
-  config.include(Devise::Test::IntegrationHelpers, type: :feature)
+  config.include(Devise::Test::IntegrationHelpers, type: :system)
   config.include(Devise::Test::ControllerHelpers, type: :controller)
+
+  config.before(:each, type: :system) do
+    driven_by(:selenium_chrome_headless)
+  end
 end
--- a/Gemfile
+++ b/Gemfile
@@ -51,7 +51,7 @@ group :development do
 end
 
 group :test do
-  gem "capybara-screenshot"
+  gem "capybara"
   gem "factory_bot_rails"
   gem "faker"
   gem "simplecov"
--- a/Gemfile
+++ b/Gemfile
@@ -37,7 +37,6 @@ group :development, :test do
   gem "rubocop-performance"
   gem "rubocop-rails"
   gem "rubocop-rspec"
-  gem "selenium-webdriver"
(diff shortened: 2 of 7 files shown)


reinisla:
Could use some help with deprecation warnings

jeppester:
> Could use some help with deprecation warnings

I found this PR regarding the deprecation warnings:
https://github.com/rails/rails/pull/43496

I think they will go away next time rails is updated.

spec/rails_helper.rb
```diff
@@ -65,6 +65,10 @@
   # config.filter_gems_from_backtrace("gem name")
 
   # Include integration helpers for feature tests
-  config.include(Devise::Test::IntegrationHelpers, type: :feature)
+  config.include(Devise::Test::IntegrationHelpers, type: :system)
   config.include(Devise::Test::ControllerHelpers, type: :controller)
+
+  config.before(:each, type: :system) do
+    driven_by(:selenium_chrome_headless)
+  end
```
jeppester, line 73: In [spec/support/capybara.rb](https://github.com/abtion/muffi/blob/main/spec/support/capybara.rb#L13) we also set the webdriver (and allow it to be overridden by an env var).

I wonder why these new three lines are needed? If they override what we have in `capybara.rb`?

runephilosof-abtion: I am uncertain, we should try it without those lines.
And if they are needed for some reason, they should use the same env var and be moved to `capybara.rb`.

runephilosof-abtion:
The deprecation warnings are:
```
WARN Selenium [DEPRECATION] [:browser_options] :options as a parameter for driver initialization is deprecated. Use :capabilities with an Array of value capabilities/options if necessary instead.
WARN Selenium [DEPRECATION] Manager#logs is deprecated. Use Chrome::Driver#logs instead.
```

The first one is discussed in the rails issue that Jesper mentioned.

spec/rails_helper.rb
reinisla: We add this so it runs headlessly (does not open chrome)

runephilosof-abtion: Yes, but the same thing is configured in `capybara.rb`.
Either way, I think it should be moved to `capybara.rb`, since it is Capybara specific configuration.

jeppester:
It turns out that rails system tests do not respect the capybara driver set in capybara's configuration:
https://github.com/rails/rails/issues/34379

I've changed the code so that we set the capybara driver the "rails way", but inside the capybara configuration where I think people will be looking for it. At least you will now have a chance to notice the code should you try to set the driver the capybara way (which would not work).

jeppester:
Btw. I have also removed capybara_screenshot, since it's made redundant by the built-in functionality of rails system tests. I have ensured that the built-in screenshot functionality works as expected.

jeppester:
> The deprecation warnings are:
> 
> ```
> WARN Selenium [DEPRECATION] [:browser_options] :options as a parameter for driver initialization is deprecated. Use :capabilities with an Array of value capabilities/options if necessary instead.
> WARN Selenium [DEPRECATION] Manager#logs is deprecated. Use Chrome::Driver#logs instead.
> ```
> 
> The first one is discussed in the rails issue that Jesper mentioned.

We also get both if these warnings in the main branch now, so they are not relevant to this PR. They will likely go away when we get new versions of rails and capybara,

runephilosof-abtion:
Awesome

runephilosof-abtion (approved):

PR #680, opened 2021-12-23
Docker asdf

jeppester:
asana:
https://app.asana.com/0/1142794766483633/1201502091102362
https://app.asana.com/0/1142794766483633/1200833935162794


hcarreras:
Cool.

I agree it's a good idea to keep the databases in docker in general. 
Something we can add (as an option) to our projects.

PR #683, opened 2021-12-23
Add rails-i18n gem with default locale translations

reinisla:


config/locales/shared/da.yml
```diff
@@ -0,0 +1,173 @@
+---
+da:
```
runephilosof-abtion, line 2: Why add the translations here when they are in the added `rails-i18n` gem?

reinisla: For most apps that I've worked on we would have to go to rails-i18n repository to lookup/fetch default translations.
If you would want to change a default translation you would need to know the translation key and then you would have to go to rails-i18n repo to look it up which would be quite annoying. 
Having translations extracted from gem makes it more convenient to edit and also helps developers to understand how we structure translations.

runephilosof-abtion: Valid arguments.
However, if some translations are improved upstream we don't get it.
So I prefer only overwriting the translations, when needed.

@jeppester What do you think?

PR #684, opened 2021-12-23
Feature/change default locale to danish

reinisla:


lib/tasks/development.rake
```diff
@@ -53,6 +53,7 @@ unless Rails.env.production?
 
   # Run the tests and fail if coverage is below 100%
   task test_strict: :environment do
+    Dotenv.overload(".env.test")
```
reinisla, line 56: Rake tasks use .env.test by default when they are called so it needs to be overloaded. I've encountered this issue in other projects and there is a task in asana [Rake task test_all_strict uses .env.development vars instead of .env.test](https://app.asana.com/0/1142794766483633/1200878922116037/f)

Is this quick-fix good enough?

reinisla, line 56: Rake tasks use .env.test by default when they are called so it needs to be overloaded. I've encountered this issue in other projects and there is a task in asana [Rake task test_all_strict uses .env.development vars instead of .env.test](https://app.asana.com/0/1142794766483633/1200878922116037/f)

Is this quick-fix good enough?

jeppester: I think this is fine for now :+1: 

IMO it exposes an underlying problem though: I think we are using rake for things that would be better solved with shell scripts.

Shell scripts replicate what people actually run in their terminals to execute the specs, whereas these rake tasks adds a layer of complexity that might cause problems like this, and just seems unnecessary to me.

.env.development
```diff
@@ -5,3 +5,4 @@ REDIS_URL=redis://localhost:6379
 SIDEKIQ_PASSWORD=password
 SIDEKIQ_USERNAME=abtion
 USE_LETTER_OPENER=true
+LOCALE=da
```
jeppester: I'm moving this to `.env` so we don't get surprises when we deploy to the production environment

PR #685, opened 2021-12-23
Feature/style devise pages similary to the login page

RyvardAbtion:
Added a shared layout for devise pages by adding a new parent controller for devise. 
Style all devise pages similar to sign in page. 

https://app.asana.com/0/1142794766483633/1200591585984903/f

Screenshot 2021-12-23 at 15 56 35

### DIFF
--- a/app/views/shared/_nav.html.erb
+++ b/app/views/shared/_nav.html.erb
@@ -3,7 +3,7 @@
     <div class="flex justify-between flex-wrap">
       <div class="flex">
         <div>
-          <%= link_to "Muffi", root_path, { class: "NavLink flex items-center py-4 px-2 font-semibold" } %>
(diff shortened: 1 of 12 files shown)


app/views/shared/_nav.html.erb
```diff
@@ -3,7 +3,7 @@
     <div class="flex justify-between flex-wrap">
       <div class="flex">
         <div>
-          <%= link_to "Care1 VCI", root_path, { class: "NavLink flex items-center py-4 px-2 font-semibold" } %>
+          <%= link_to "Muffi", root_path, { class: "NavLink flex items-center py-4 px-2 font-semibold" } %>
```
jeppester, line 6: Well spotted! We should change it to "Project Name Human"-though

jeppester:
I would like us to have some tests for this, but I'm not sure that I think unit testing the views is the right approach.

I'm merging this for now, and adding a card to the muffi board regarding testing of devices features.

PR #686, opened 2021-12-23
Feature/sendinblue setup

reinisla:
Account has been set up.
However it seems that sendinblue is not sending anything.
The remaining mail amount is decreasing, but I haven't received any mails.
Could be that there are some additional steps that need to be made, like [account activation](https://help.sendinblue.com/hc/en-us/articles/209463245)
Or maybe we just need to upgrade the plan to paid one

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -281,18 +281,12 @@ GEM
     rubocop-rspec (2.6.0)
       rubocop (~> 1.19)
     ruby-progressbar (1.11.0)
-    ruby_http_client (3.5.2)
     rubyzip (2.3.2)
     selenium-webdriver (4.1.0)
       childprocess (>= 0.5, < 5.0)
       rexml (~> 3.2, >= 3.2.5)
       rubyzip (>= 1.2.2)
     semantic_range (3.0.0)
-    sendgrid-actionmailer (3.2.0)
-      mail (~> 2.7)
-      sendgrid-ruby (~> 6.4)
-    sendgrid-ruby (6.6.0)
-      ruby_http_client (~> 3.4)
     shoulda-matchers (5.1.0)
       activesupport (>= 5.2.0)
     sidekiq (6.3.1)
@@ -384,7 +378,6 @@ DEPENDENCIES
   rubocop-performance
   rubocop-rails
   rubocop-rspec
-  sendgrid-actionmailer
   shoulda-matchers
   sidekiq
   simplecov
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -135,6 +135,9 @@ GEM
       activesupport (>= 5.0)
     highline (2.0.3)
     html_tokenizer (0.0.7)
+    httparty (0.20.0)
+      mime-types (~> 3.0)
+      multi_xml (>= 0.5.2)
     i18n (1.8.11)
       concurrent-ruby (~> 1.0)
     i18n-tasks (0.9.36)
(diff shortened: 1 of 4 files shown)


jeppester:
I wonder how you set up the DNS settings?

Usually we have a domain/subdomain for the app, and we need to set a bunch of DNS records for that domain in order for emails to not end up in spam. In this case I don't think we have a muffi domain at all.

config/environments/production.rb
```diff
@@ -55,15 +55,19 @@
   # config.cache_store = :mem_cache_store
 
   # Use a real queuing backend for Active Job (and separate queues per environment).
-  # config.active_job.queue_adapter     = :resque
+  config.active_job.queue_adapter = :sidekiq
```
jeppester, line 58: Was this necessary?

It seems like we have code that does the same in `config/initializers/active_job.rb`.

README.md
```diff
@@ -161,12 +161,18 @@ If you need to ignore specific translation keys, follow this process:
...
+   2. Email: Client google group account
+   3. Password: Generate in 1Password
+2. You must validate a phone number in order to start sending mails.
+3. Generate SENDINBLUE keys
+   1. Go to SMTP & API under user settings
+   2. Click on SMTP Tab
+   3. Create a new SMTP KEY.
+   4. Copy smtp key value to `SENDINBLUE_PASSWORD` variable on clients heroku project
+   5. Copy login to `SENDINBLUE_USERNAME` variable on clients heroku project
+4. If you want the review apps to send emails: Copy the staging app's Sendinblue credentials and add them to the review apps config vars.
```
jeppester, line 175: I really like this, great work :+1:

README.md
```diff
@@ -161,12 +161,18 @@ If you need to ignore specific translation keys, follow this process:
...
-2. Add Sendgrid addon.
-3. Click the added Sendgrid resource, go through the confirmation steps.
-4. Set up DKIM on the client's domain (under sender authentication -> domain authentication).
-5. Go to staging and repeat the above steps.
-6. If you want the review apps to send emails: Copy the staging app's Sendgrid credentials and add them to the review apps config vars. Do not add the Sendgrid addon to review apps in `app.json` (Sendgrid will ban our account).
+1. Register and add to 1Password new sendinblue account
+   1. Register: https://app.sendinblue.com/account/register
+   2. Email: Client google group account
+   3. Password: Generate in 1Password
+2. You must validate a phone number in order to start sending mails.
```
jeppester, line 168: I wonder if we can standardize this more, e.g. have a specific phone that we always use.

jeppester: @hcarreras What is your input on this?

jeppester:
I'm merging this. I think it's more than good enough and we can always improve.

config/environments/production.rb
reinisla: It is not necessary, I added this thinking that it might be sidekiq deliver later issue that mails were not sending.

reinisla:
@jeppester Do you think that DNS records need to be setup always? In sendgrid you did not need to, for it to send mails. I understand the getting through spam reasoning but the mails simply did not send at all. 

What I meant with account activation is more related to this point in the url that I commented above: _"Create a ticket or contact us here to request the account activation."_

Gemfile.lock
```diff
@@ -384,7 +389,7 @@ DEPENDENCIES
   rubocop-performance
   rubocop-rails
   rubocop-rspec
-  sendgrid-actionmailer
+  sendinblue
```
runephilosof-abtion, line 392: We aren't using the sendinblue gem for anything.
So I suggest removing it.

README.md
hcarreras: Totally agree. I'll ask Angelle to get us a shared phone number

PR #709, opened 2022-01-04
Upgrade to rails 7

jeppester:
Asana: https://app.asana.com/0/1142794766483633/1201610634245093


config/application.rb
```diff
@@ -2,20 +2,7 @@
...
-require "active_job/railtie"
-require "active_record/railtie"
-require "active_storage/engine"
-require "action_controller/railtie"
-require "action_mailer/railtie"
-require "action_mailbox/engine"
-require "action_text/engine"
-require "action_view/railtie"
-require "action_cable/engine"
-# require "sprockets/railtie"
```
runephilosof-abtion, line 17: :+1: 
Sprockets are not a part of `rails/all` anymore https://github.com/rails/rails/blob/main/railties/lib/rails/all.rb

config/application.rb
```diff
@@ -2,20 +2,7 @@
...
-# require "rails/test_unit/railtie"
```
runephilosof-abtion, line 18: `test_unit` is a part of `rails/all`, but I guess it is not a problem that it is loaded?

config/initializers/content_security_policy.rb
```diff
@@ -5,30 +5,23 @@
...
-# Rails.application.config.content_security_policy do |policy|
-#   policy.default_src :self, :https
-#   policy.font_src    :self, :https, :data
-#   policy.img_src     :self, :https, :data
-#   policy.object_src  :none
-#   policy.script_src  :self, :https
-#   policy.style_src   :self, :https
-#   # If you are using webpack-dev-server then specify webpack-dev-server host
-#   if Rails.env.development?
-#     policy.connect_src :self, :https, "http://localhost:3035", "ws://localhost:3035"
```
runephilosof-abtion, line 17: I think this was removed by accident.
We want this to avoid problems with webpack-dev-server in development.

config/application.rb
jeppester: I ran the upgrade script, it didn't complain about conflicts in application.rb.

I'll change it back so we only get the parts we need.

config/initializers/content_security_policy.rb
jeppester: Since this was all commented out, I thought we hadn't made changes to the defaults.

I'll re-add the webpack-dev-server-part, and have a look in the git logs to see if we made other changes to the file that we want to keep.

jeppester: It turns out that the webpacker-specific code has been moved from rails to webpacker:
https://github.com/rails/rails/commit/82e44320589fcbacf16012fd7fdca07835b88f20#diff-a8088f5673c46f1598d74d8502f881f005063142132fe7e42d6ae1d2c6bca190L14
https://github.com/rails/webpacker/blob/550c53eb7816140a24b3af9691fbbdf23f24a563/lib/install/template.rb#L57

Unfortunately it seems to have the side-effect that the rails upgrade script no longer includes it.

jeppester: This has now been fixed in main

config/application.rb
jeppester: Fixed in main

PR #742, opened 2022-01-24
Simplify scripts

jeppester:
Asana: https://app.asana.com/0/1142794766483633/1201686018033960


bin/test-failfast
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+set -e
+
+bundle exec rspec --order defined --fail-fast $(find spec -iname "*_spec.rb" -exec ls -1t "{}" +;)
```
jeppester, line 4: I extracted this script from development.rake, but while working on it I realized that it wasn't working in the muffi template (It called "bin/rspec" which we don't have in muffi)

It made me wonder if anyone is using this feature at all.
For now I've decided to keep it because I think it was the least controversial decision

I think we should revisit - and possibly remove - it at some point, I'll at it to the muffi board.

bin/pre-commit
```diff
@@ -3,3 +3,4 @@
 echo "Running pre-commit hook"
 
 ./bin/rspec-rubocop-staged-files
+yarn test --onlyChanged
```
runephilosof-abtion, line 6: This seems to be a new feature that snuck itself into this refactor, right?

bin/test-failfast
MarkusAbtion: I remember it being very useful on Montus, because it would run the test suite sorted by last failing tests, and if a test still failed, it wouldn't continue running.

PR #928, opened 2022-05-03
Default capybara browser size

reinisla:
Since we removed configs for the default size of browser [here](https://github.com/abtion/muffi/commit/1dd08422c754b7c7a4339f2a9817b3f679d3daea#diff-d522976f73ce246aff733a18173712d72bb311e61561ea37eb1ee0d04b78fe9aL31 ) 
It is causing some issues when updating capybara configs in other projects


reinisla:
@jeppester @runephilosof-abtion Any alternatives?

PR #955, opened 2022-05-17
Chore/ENVs to constants

reinisla:
Asana: https://app.asana.com/0/1142794766483633/1200395882092323/f

This task turned out to be messier than expected.
The example project did not include ENV's fetched in initialisers, so we had to come up with a solution.

There are still few ENV.fetch calls in places ~where we could not extract them from due to loading order~

- ~environment/development.rb~ , 
- ~environment/production.rb~
- database.yml
- cable.yml
- puma.rb
- ~code_coverage_helper.rb~

So not entirely sure what we are supposed to do with those

~There is a thing called [preinitilizer](https://willbryant.net/rails_initialization_and_configuration_order) configs but that would require us to move constants out of initializer folder so that they dont get initialized twice.~


config/initializers/01_constants.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Constants
+  SEED_ADMIN_EMAIL = ENV.fetch("SEED_ADMIN_EMAIL", "admin@example.com")
```
runephilosof-abtion, line 4: I would prefer to use `fetch` without fallback and keep the defaults in `.env.development`.
That way the app will fail in production if they are missing.

reinisla:
Had a small pairing session with @jeppester and we made a change to move/rename `initializers/01_constants.rb` to `config/constants.rb`
That way we can require constants in `application.rb` and `spec_helper`, this allows ENVs in `config/environments/` to also use constants.

There is still a question of: do we want to constantize all of the env fetch calls? Files like `puma.rb` would have to require the constants manually due to the `procfile` web script that only calls the `puma.rb`.

I would suggest that any environmental variable fetch in `initializers` and `environments` should be constantized

.env.development
```diff
@@ -1,7 +1,7 @@
 SEED_ADMIN_EMAIL=admin@example.com
-SEED_ADMIN_INITIAL_PASSWORD=password
+SEED_ADMIN_INITIAL_PASSWORD=F8osN3hiRjJ0dstGJUDR
```
runephilosof-abtion, line 2: I prefer something easier to type for the cases where you want to test on a device without access to copy paste from the file
```suggestion
SEED_ADMIN_INITIAL_PASSWORD=`liga guide prep mare`
```

Although, I think the decision at our last Muffi meeting was to not change the password...

config/constants.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Constants
+  SEED_ADMIN_EMAIL = ENV.fetch("SEED_ADMIN_EMAIL", nil)
+  SEED_ADMIN_INITIAL_PASSWORD = ENV.fetch("SEED_ADMIN_INITIAL_PASSWORD", nil)
+  MAIL_FROM = ENV.fetch("MAIL_FROM", nil)
```
runephilosof-abtion, line 6: A lot of these should be mandatory, like so

```suggestion
  MAIL_FROM = ENV.fetch("MAIL_FROM")
```

config/constants.rb
```diff
@@ -0,0 +1,31 @@
...
+  SIDEKIQ_USERNAME = ENV.fetch("SIDEKIQ_USERNAME", nil)
+  SIDEKIQ_PASSWORD = ENV.fetch("SIDEKIQ_PASSWORD", nil)
```
runephilosof-abtion, line 8: At least those that were mandatory before should still be mandatory

```suggestion
  SIDEKIQ_USERNAME = ENV.fetch("SIDEKIQ_USERNAME")
  SIDEKIQ_PASSWORD = ENV.fetch("SIDEKIQ_PASSWORD")
```

config/constants.rb
```diff
@@ -0,0 +1,31 @@
...
+  BLOCKED_IPS = ENV.fetch("BLOCKED_IPS", nil)
+  BACKTRACE = ENV.fetch("BACKTRACE", nil)
+  RAILS_SERVE_STATIC_FILES = ENV.fetch("RAILS_SERVE_STATIC_FILES", nil)
+  USE_LETTER_OPENER = ENV.fetch("USE_LETTER_OPENER", nil)
+  ALLOWED_HOSTS = ENV.fetch("ALLOWED_HOSTS", "")
+  RAILS_LOG_TO_STDOUT = ENV.fetch("RAILS_LOG_TO_STDOUT", nil)
+  SENDINBLUE_EMAIL = ENV.fetch("SENDINBLUE_EMAIL", nil)
+  SENDINBLUE_PASSWORD = ENV.fetch("SENDINBLUE_PASSWORD", nil)
+  DOMAIN_NAME = ENV.fetch("DOMAIN_NAME", nil)
+  HEROKU_APP_NAME = "#{ENV.fetch('HEROKU_APP_NAME', nil)}.herokuapp.com".freeze
```
runephilosof-abtion, line 28:
```suggestion
  HEROKU_APP_NAME = "#{ENV.fetch('HEROKU_APP_NAME')}.herokuapp.com".freeze
```

config/constants.rb
```diff
@@ -0,0 +1,31 @@
...
+  SEED_ADMIN_INITIAL_PASSWORD = ENV.fetch("SEED_ADMIN_INITIAL_PASSWORD", nil)
+  MAIL_FROM = ENV.fetch("MAIL_FROM", nil)
+  SIDEKIQ_USERNAME = ENV.fetch("SIDEKIQ_USERNAME", nil)
+  SIDEKIQ_PASSWORD = ENV.fetch("SIDEKIQ_PASSWORD", nil)
+  HTTP_AUTH_USERNAME = ENV.fetch("HTTP_AUTH_USERNAME", nil)
+  HTTP_AUTH_PASSWORD = ENV.fetch("HTTP_AUTH_PASSWORD", nil)
+  LOCALE = ENV.fetch("LOCALE", "en")
+  ROLLBAR_ACCESS_TOKEN = ENV.fetch("ROLLBAR_ACCESS_TOKEN", nil)
+  ROLLBAR_ENV = ENV.fetch("ROLLBAR_ENV", nil)
+  REDIS_URL = ENV.fetch("REDIS_URL", nil)
```
runephilosof-abtion, line 14:
```suggestion
  REDIS_URL = ENV.fetch("REDIS_URL")
```

.env.development
jeppester: Let keep "password"

config/constants.rb
jeppester: HEROKU_APP_NAME is something heroku provides for review apps.
In the production config we have some logic around for setting the host name to HEROKU_APP_NAME only if it's available. Otherwise we use the DOMAIN_NAME var.

It is a mistake that the string templating was moved to `constants.rb` however, since it means that the app will always have a HEROKU_APP_NAME, which is not the purpose.

The question is whether or not we want to keep this functionality around when we are not currently using review apps AFAIK on any recent projects.

.env.development
reinisla: due to pwned password protection we cant use "password" anymore, so i'm ok with using random word group instead

config/constants.rb
hcarreras: I believe `HEROKU_APP_NAME` is given to all apps https://devcenter.heroku.com/articles/dyno-metadata
I vaguely remember seeing an addon using it (can't remember whether it would work if the name was nullified).

I agree that it is better to reduce name usage to `DOMAIN_NAME` unless we have a good reason. This will reduce possible differences between the value of `HEROKU_APP_NAME` in Heroku environment and the app environment.

hcarreras:
First of all, great job 👏 I've been looking forward to this one for a while. It's a small improvement, but we have had many issues due to poor handling of required ENV variables. This makes the ENV variables highly explicit, so it gets harder to miss.

If I understand the issue correctly, `puma.rb` would need to `require "constants"`? 
Why is that an issue? If that can move all ENV variables to the same place, I'd say it's worth it.

Am I missing something?

reinisla:
@hcarreras The issue is that it is not only `puma.rb` that would need `require "constants"`
These files would also need the require:
`config/`
- `boot.rb`
- `cable.yml`
- `database.yml`

Also binstubs seems to be using ENVs

.env.development
reinisla: I've updated it to be word string, however if we add strong password validation it wont pass due to lack of numbers/uppercase letters

config/constants.rb
reinisla: Not sure what is the conclusion here, so I've kept it as it is. Do we want to remove it from `constants.rb` AND `production.rb` ?

.env.development
```diff
@@ -1,7 +1,7 @@
 SEED_ADMIN_EMAIL=admin@example.com
-SEED_ADMIN_INITIAL_PASSWORD=password
+SEED_ADMIN_INITIAL_PASSWORD=muffi hero dog
 MAIL_FROM=mail_from@example.com
 REDIS_URL=redis://localhost:6379
-SIDEKIQ_PASSWORD=password
 SIDEKIQ_USERNAME=abtion
+SIDEKIQ_PASSWORD=muffi hero dog
```
MarkusAbtion, line 6: Do env-values with spaces work without quotes?

jeppester:
We will attempt to replace this with something else.

PR #956, opened 2022-05-19
Feature/devise pwned password protection

reinisla:
For new user creation, however we do not enable registrations on muffi so this is only relevant when creating users via console.

Update:
Added after sign in flash warning


runephilosof-abtion:
:+1:

PR #967, opened 2022-05-25
Feature/admin area

reinisla:
Admin area implementation with user management section

Some of the elements are using Link, Button components

To access admin area user must be admin: true
There is no link in navigation to admin area therefore you must go to /admin/users to access it

Screen Recording 2022-05-25 at 11 57 26 AM gif

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -108,6 +108,7 @@ RSpec/MultipleExpectations:
   Max: 5
   Exclude:
     - "spec/mailers/**/*"
+    - "spec/views/**/*"
 
 RSpec/LeadingSubject:
   Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -32,7 +32,6 @@ Lint/ScriptPermission:
     - "Rakefile"
 
 Metrics/AbcSize:
-  Max: 15
   Exclude:
     - "spec/**/*"
     - "db/migrate/*.rb"
--- a/app/controllers/admin/base_controller.rb
+++ b/app/controllers/admin/base_controller.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module Admin
+  class BaseController < ApplicationController
+    before_action :validate_admin
+
+    private
+
+    def validate_admin
+      authorize(current_user, :admin?)
+    end
+  end
+end
--- a/app/controllers/admin/base_controller.rb
+++ b/app/controllers/admin/base_controller.rb
@@ -2,7 +2,7 @@
 
 module Admin
   class BaseController < ApplicationController
-    after_action :verify_authorized, except: :index
+    after_action :verify_authorized
     after_action :verify_policy_scoped, only: :index
 
     def index; end
--- a/app/controllers/admin/base_controller.rb
+++ b/app/controllers/admin/base_controller.rb
@@ -2,12 +2,9 @@
 
 module Admin
   class BaseController < ApplicationController
-    before_action :validate_admin
+    after_action :verify_authorized, except: :index
+    after_action :verify_policy_scoped, only: :index
 
-    private
-
-    def validate_admin
-      authorize(current_user, :admin?)
-    end
+    def index; end
   end
 end
--- a/app/javascript/components/Button/index.scss
+++ b/app/javascript/components/Button/index.scss
@@ -31,7 +31,8 @@
   }
 
   // Colors
-  &:enabled {
+  &:enabled,
+  &.Button__link {
     &.Button--neutral {
       @apply text-neutral-600-contrast bg-neutral-500 hover:bg-neutral-600 active:bg-neutral-700 ring-neutral-500;
     }
--- a/app/javascript/components/Button/index.scss
+++ b/app/javascript/components/Button/index.scss
@@ -31,8 +31,9 @@
   }
 
   // Colors
-  &:enabled,
-  &.Button__link {
+  // We make a negative check on disabled because it allows this styling to work with tags
+  // that don't have enabled/disabled states
+  &:not(:disabled) {
     &.Button--neutral {
       @apply text-neutral-600-contrast bg-neutral-500 hover:bg-neutral-600 active:bg-neutral-700 ring-neutral-500;
     }
--- a/app/javascript/components/Link/index.tsx
+++ b/app/javascript/components/Link/index.tsx
@@ -16,9 +16,9 @@ export enum LinkVariant {
 }
 
 export interface LinkProps extends React.HTMLAttributes<HTMLAnchorElement> {
-  variant: LinkVariant
+  variant?: LinkVariant
   text: string
-  size: LinkSize
+  size?: LinkSize
   underline?: boolean
 }
 
--- a/app/javascript/components/Link/index.tsx
+++ b/app/javascript/components/Link/index.tsx
@@ -1,0 +1,39 @@
+import React from "react"
+import classNames from "classnames"
+import "./index.scss"
+
+export enum LinkSize {
+  ExtraSmall = "xs",
+  Small = "sm",
+  Base = "bs",
+}
+
+export enum LinkVariant {
+  Neutral = "neutral",
+  NeutralLight = "neutral-light",
+  NeutralDark = "neutral-dark",
+  Info = "info",
+}
+
+export interface LinkProps extends React.HTMLAttributes<HTMLAnchorElement> {
+  variant: LinkVariant
+  text: string
+  size: LinkSize
+  underline?: boolean
+}
+
+export default function Link(props: LinkProps): JSX.Element {
+  const { underline, size, className, variant, text, ...rest } = props
+
+  const usedClassName = classNames("Link", className, {
+    [`Link--${variant}`]: variant,
+    [`Link--${size}`]: size,
+    [`Link--underline`]: underline,
+  })
+
+  return (
+    <a className={usedClassName} {...rest}>
+      {text}
+    </a>
+  )
+}
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -13,34 +13,28 @@ class ApplicationPolicy
     @record = record
   end
 
-  delegate :admin?, to: :user
-
-  def index?
-    admin?
-  end
-
   def show?
-    admin?
+    user.admin?
   end
 
   def create?
-    admin?
+    user.admin?
   end
 
   def new?
-    admin?
+    user.admin?
   end
 
   def update?
-    admin?
+    user.admin?
   end
 
   def edit?
-    admin?
+    user.admin?
   end
 
   def destroy?
-    admin?
+    user.admin?
   end
 
   class Scope
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -13,32 +13,34 @@ class ApplicationPolicy
     @record = record
   end
 
+  delegate :admin?, to: :user
+
   def index?
-    user.admin?
+    admin?
   end
 
   def show?
-    user.admin?
+    admin?
   end
 
   def create?
-    user.admin?
+    admin?
   end
 
   def new?
-    create?
+    admin?
   end
 
   def update?
-    user.admin?
+    admin?
   end
 
   def edit?
-    update?
+    admin?
   end
 
   def destroy?
-    user.admin?
+    admin?
   end
 
   class Scope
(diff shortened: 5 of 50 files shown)


.rubocop.yml
```diff
@@ -108,6 +108,7 @@ RSpec/MultipleExpectations:
   Max: 5
   Exclude:
     - "spec/mailers/**/*"
+    - "spec/views/**/*"
```
reinisla, line 111: Mailers and views usually have a lot of expectations so I think it makes sense to disable it there. Otherwise we would need start using `        expect(rendered_component).to eq(
          <<~HTML ...content.. HTML`

app/controllers/admin/base_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Admin
+  class BaseController < ApplicationController
+    before_action :validate_admin
+
+    private
+
+    def validate_admin
+      authorize(current_user, :admin?)
```
reinisla, line 10: Since we have pundit I've decided to use it here

app/javascript/components/Button/index.scss
```diff
@@ -31,7 +31,8 @@
   }
 
   // Colors
-  &:enabled {
+  &:enabled,
+  &.Button__link {
```
reinisla, line 35: Current button components dont support hrefs, so I've added this class make it work with not only button elements

app/javascript/components/Link/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+  variant: LinkVariant
+  text: string
+  size: LinkSize
+  underline?: boolean
+}
+
+export default function Link(props: LinkProps): JSX.Element {
+  const { underline, size, className, variant, text, ...rest } = props
+
+  const usedClassName = classNames("Link", className, {
```
reinisla, line 28: I've added `className` option, in muffibook example it is not supported. This adds possibility to pass additional classnames to react_component

app/views/admin/users/_form.html.erb
```diff
@@ -0,0 +1,22 @@
...
+    <%= f.label(:password, class: "inline-block mb-2") %>
+    <%= f.password_field(:password, autocomplete: "new-password", placeholder: true, class: "Input Input--neutral Input--md block w-full") %>
+  </div>
+  <div class="mb-12">
+    <%= f.label(:password_confirmation, class: "inline-block mb-2") %>
+    <%= f.password_field(:password_confirmation, autocomplete: "new-password", placeholder: true, class: "Input Input--neutral Input--md block w-full") %>
+  </div>
+  <div class="mb-6">
+    <%= react_component :Button, className: "w-full mb-2", size: "lg", variant: :primary, type: :submit, children: t("admin.shared.save") %>
+    <a role="button" href="<%= admin_users_path %>" class="Button Button__link Button--lg Button--neutral inline-block text-center w-full"><%= t("admin.shared.cancel") %></a>
```
reinisla, line 20: Example of "link like button"

app/views/admin/users/index.html.erb
```diff
@@ -0,0 +1,50 @@
...
+                  </td>
+                </tr>
+              <% end %>
+            </tbody>
+          </table>
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
```
reinisla, line 50: Tailwind table styling taken from https://tailwind-elements.com/docs/standard/components/tables/

app/views/devise/registrations/edit.html.erb
```diff
@@ -38,7 +38,7 @@
     <small class="text-sm block text-right">
       <p><%= t(".unhappy") %> <%= link_to t(".cancel_my_account"), registration_path(resource_name), data: { confirm: t(".are_you_sure") }, method: :delete %></p>
 
-      <%= link_to t("devise.shared.links.back"), :back %>
+      <%= link_to t("devise.shared.links.back"), :back, class: "Link Link--info Link--bs inline-block" %>
```
reinisla, line 41: I'd assume we would want to replace all link_to with react_component "Link"

spec/views/admin/users/index.html.erb_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    expect(rendered).to have_css("tr th", text: "Email")
+    expect(rendered).to have_css("tr td", text: user.email.to_s)
+
+    expect(rendered).to have_css("tr th", text: "Name")
+    expect(rendered).to have_css("tr td", text: user.name.to_s)
+
+    expect(rendered).to have_css("tr th", text: "Created at")
+    expect(rendered).to have_css("tr td", text: user.created_at.strftime("%d %b %H:%M").to_s)
+    expect(rendered).to have_css("[data-react-component=Link]", count: 2)
```
reinisla, line 25: Not sure what is the best way to test react_components in view specs

app/views/devise/registrations/edit.html.erb
jeppester: As long as we can get along just fine using the classes of the react components, that's perfectly fine.
I would actually say it's preferable since it's easier to test in view specs for instance.

This is the reason that we are using BEM over scoped styling / styled components / etc. - it allows us to just use the component styling, which is mostly enough for buttons / inputs / alerts etc.

app/javascript/components/Button/index.scss
jeppester: Good point. I think we should update the button component to support links, it's a simple thing I know I've implemented on multiple projects.

app/javascript/components/Link/index.tsx
jeppester: Great idea :+1:, the components should be as flexible as possible.

app/controllers/admin/base_controller.rb
jeppester: I think the point of the application policy is to allow us to just have an `authorize` call in each of the admin controllers, without having to create policies for each resource (or maybe we need to, but each of them could just be empty and extending `ApplicationPolicy`).

I'm not sure if I would like to do that instead of what you do here. This has the benefit of keeping the controllers more simple.

app/views/devise/registrations/edit.html.erb
jeppester: However: All this "class spam" makes me wonder if it even makes sense to have a Link component in Muffibook.

All of this could very likely have just been:
```suggestion
      <%= link_to t("devise.shared.links.back"), :back, class: "text-info inline-block" %>
```

I mean, the point of the BEM-classes is to make the more elaborate styling easier to implement. If we instead make simple styling harder to implement, I think we are not doing it right.

I'll take this up with Jacob.

spec/views/admin/users/index.html.erb_spec.rb
jeppester: I think this is the best we can do without going down the rabbit hole of checking all props etc.
However: We could have just used the BEM-classes for these links instead of `react_component`, and in that case it would have been easier to test.

jeppester:
Lots of decisions had to be taken here. I want to thank you for taking this on and putting in such an effort! This was never going to be one of the quick and easy ones. It's also the first time we've taken our own muffibook medicine and faced it's quirks and rough edges.

There are a few things I would like to improve:
- I think we should generally only use `react_component` when it's really necessary (for things that need state - like the dismissable alerts - I'll make a task to clarify somewhere)
- I would like a designer's eye on the tables (these might end up getting copied a LOT in projects)
- I'm not entirely sure what to think about the current authorization for admin. It's simple, but it also feels like we are working around how pundit is meant to be implemented.

And there are a couple of things we'll need to figure out how to solve in muffibook:
- Should we have a link component or is it overkill?
- Buttons should have a `LinkTo` prop, so that they can be links.

At the same time, this is "good enough" for a start, so I don't think it would be a problem to merge it.
Still I suggest that pair on this soon to get it past the finish line.

app/javascript/components/Link/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export interface LinkProps extends React.HTMLAttributes<HTMLAnchorElement> {
+  variant: LinkVariant
+  text: string
+  size: LinkSize
+  underline?: boolean
+}
+
+export default function Link(props: LinkProps): JSX.Element {
```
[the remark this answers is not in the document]
reinisla: It is here https://github.com/abtion/muffi/pull/967/files#diff-1a37e8d8224a1fd673f4d0348683596e4f6b77dd3e4380b5d32b1a504ab8148eR40, however @jeppester is suggesting that maybe we shouldnt have a react component for Links https://github.com/abtion/muffi/pull/967#discussion_r885238173 and from using them in this project I would agree, it is quite painful process to use it

app/controllers/admin/base_controller.rb
reinisla: I would normally authorize each action in controller, however in this case https://github.com/abtion/muffi/blob/main/app/policies/application_policy.rb is set up to check for `user.admin?` in the base level and is not really used, so I'm not sure how to proceed we also only have one "role".

jeppester: I've updated the PR to authorize on each action - except for index which uses policy scope.
That seems to be the intended usage of pundit judging from their documentation.

spec/views/admin/users/index.html.erb_spec.rb
jeppester: I've replaced the react components for these links with plain links using the relevant "Link" classes

app/javascript/components/Button/index.scss
jeppester: I've fixed this issue by changing `&:enabled` to `&:not(:disabled)`.
With that change we don't need an additional `Button__link` class to make links work with the button classes.

app/javascript/components/Link/index.tsx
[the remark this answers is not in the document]
jeppester: After having given this a some extra thought I guess I think the link component is fine to have - for the sake of consistency.

But we should make it clear in the muffibook examples where the styling originates from, so that projects can choose to just use tailwind classes instead.

For Muffi I don't think we need link-component - or we can make a normal `a`-tag behave like it's meant to according to the design.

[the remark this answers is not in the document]
jeppester: I decided to slightly alter the link-component so that it defaults to the most commonly used color.
Then the only class we need to add is "Link", and I think that's fine.

app/policies/application_policy.rb
```diff
@@ -13,10 +13,6 @@ def initialize(user, record)
     @record = record
   end
 
-  def index?
-    user.admin?
-  end
-
```
runephilosof-abtion, line 19: Why remove this?
It seems a bit out of place in this PR

app/policies/application_policy.rb
```diff
@@ -26,15 +22,15 @@ def create?
...
   end
 
   def update?
     user.admin?
   end
 
   def edit?
-    update?
+    user.admin?
   end
```
runephilosof-abtion, line 34: These changes also seem out of place.

runephilosof-abtion: It also means that the UserPolicy will return false for `edit?`, because it only overrides `update?`

jeppester: Well spotted. These changes were not intentional.

app/policies/application_policy.rb
jeppester: I changed this because I added `verify_authorized` and `verify_policy scoped`, following the second example here:
https://github.com/varvet/pundit/tree/d7c2e6e3bc1f5526f220a729090249635a7453d8#ensuring-policies-and-scopes-are-used

As the example skips the index for `verify_authorized`, I got the impression that it was not preferable to both authorize and scope for indexes.

I realize now that not having a policy method for `index` will cause some issues though. It will be difficult to create conditional links in the navigation, and it's also not great UX if you are technically able to see an index page but with nothing in it - even if there's no link for the page.

How would you feel about having something like the following instead?:
```ruby
def index?
  Scope.new(user, record) != record.none
end
```

It would allow us to have the logic in the scope, instead of spread across scope and `index?`.

jeppester: After having given this some more thought I think the solution I proposed might be in the "too clever" territory. For now I think it's best to keep things simple and easy to understand.

jeppester:
I believe this is good enough now.

Here's a quick gif of it in action (don't mind the poor compression on the background, it IS solid grey):
output

jeppester:
I'm merging this, we can easily implement eventual feedback later.

PR #968, opened 2022-05-25
Experimental feature/form error handling

reinisla:
Since we are not displaying any field errors here is something similar to what I've used before in other projects

Image 2022-05-25 at 12 52 13 PM png

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -7,7 +7,7 @@ module ApplicationHelper
     errors_array = resource.errors.messages[field]
     translated_attribute = resource.class.human_attribute_name(field)
 
-    tag.div(class: "mt-1") do
+    tag.div(class: "error-messages") do
       errors_array.map do |error|
         concat(tag.p("#{translated_attribute} #{error}.", class: "text-danger-800 text-sm"))
       end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,4 +1,16 @@
 # frozen_string_literal: true
 
 module ApplicationHelper
+  def error_message(field:, resource:)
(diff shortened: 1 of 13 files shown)


app/helpers/application_helper.rb
```diff
@@ -1,4 +1,16 @@
 # frozen_string_literal: true
 
 module ApplicationHelper
+  def error_message(field:, resource:)
```
jeppester, line 4: I think `error_message` is a too generic name for what this does.

Maybe we could rename it to `field_error`?

app/javascript/components/Input/index.scss
```diff
@@ -38,4 +38,8 @@
   &--info {
     @apply border-info-300 focus:ring-info-300 focus:border-info-300;
   }
+
+  &.field_with_errors {
```
jeppester, line 42:
```suggestion
  &--error {
```

config/initializers/errors.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+ActionView::Base.field_error_proc = proc do |html_tag, _instance|
+  if /<(input|textarea|select)/.match?(html_tag)
+    html_field = Nokogiri::HTML::DocumentFragment.parse(html_tag)
+    html_field.children.add_class("field_with_errors")
```
jeppester, line 6:
```suggestion
    html_field.children.add_class("Input--error")
```

jeppester:
This is a very good starting point! I would like to refine it a little bit before merging.
I'll move the card you added in muffi to the backlog so we can get this done.

jeppester:
I'll rebase this improvement on the main branch.
Then we can get this merged without having to wait for the user admin PR - which will likely take a little while longer (though I'm working on it)

PR #979, opened 2022-05-30
Handle routing errors

viesii:
part of

https://app.asana.com/0/1199937586489373/1202291573664006 and https://app.asana.com/0/1142794766483633/1201225530897124

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -6,6 +6,14 @@ class ErrorsController < ActionController::Base
 
   def not_found
     render status: :not_found
+
+    return unless request.referer
+
+    Rollbar.warning(
+      ["RoutingError: No route matches",
+       "[#{request.method}]",
+       "'#{request.original_url}'"].join(" ")
+    )
   end
 end
 # rubocop:enable Rails/ApplicationController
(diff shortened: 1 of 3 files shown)


app/controllers/errors_controller.rb
```diff
@@ -6,6 +6,14 @@ class ErrorsController < ActionController::Base
...
   def not_found
     render status: :not_found
+
+    return unless request.referer
+
+    Rollbar.warning(
+      ["RoutingError: No route matches",
+       "[#{request.method}]",
+       "'#{request.original_url}'"].join(" ")
+    )
```
jeppester, line 16: Do you know if `Rollbar.warning` automatically includes the stacktrace/request headers/etc.?

I think those will be very useful for eventually debugging dead links.

jeppester (approved):
I really like this, simple and easy to understand :+1:

app/controllers/errors_controller.rb
jeppester: After a quick investigation we've discovered that request headers ARE sent to rollbar. So we'll have all the info we need to debug dead links :partying_face:

MartinAbtion: ah perfect! 💪 can you send a screenshot or something so i know where to look next time?

hcarreras (commented):
Love it 👏

PR #1135, opened 2022-08-23
Run only rspec tests with coverage on CI as part of the rspec job.

alexburke-abtion:
Previously the bin/test-coverage script was called, but this includes
running the JS tests with coverage enabled which duplicates work
already done by the jest job.

Note: care was taken to keep coverage percentage defined only once

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -231,7 +231,7 @@ jobs:
           bin/rake db:create
           bin/rake db:schema:load
       - name: rspec
-        run: bin/rspec
+        run: bin/rspec-coverage
       - name: Archive capybara artifacts
         uses: actions/upload-artifact@v3
         if: failure()
--- a/bin/rspec-coverage
+++ b/bin/rspec-coverage
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+set -e
+
+CODE_COVERAGE_PERCENTAGE=100 bin/rspec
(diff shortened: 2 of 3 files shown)


jeppester:
From the description of this PR I was getting the impression the diff would include replacing `bin/test-coverage` in the CI-script.

But from what I can see it's `bin/rspec` (which does not run the FE tests) that has been replaced with a new `bin/rspec-coverage` - which effectively does the same (since `CODE_COVERAGE_PERCENTAGE` is set to 100 in `.env.test`).

It makes me wonder if this change was relevant on a client project, but not on muffi? Otherwise there must be something else I'm misunderstanding here.

Can you elaborate on how the previous code was running the JS tests twice? - or maybe that's not what you meant with duplicate work?

runephilosof-abtion:
The `test-coverage` script is only called in the pre-push git hook, not running on CI.

I wonder why `CODE_COVERAGE_PERCENTAGE` is defined anywhere else than in `.env.test`.

jeppester:
> I wonder why `CODE_COVERAGE_PERCENTAGE` is defined anywhere else than in `.env.test`.

I think it's a leftover, I'll remove it and see if it makes a difference.

@alexburke-abtion Did you get around to read the feedback I posted? I'm a bit unsure about what to do with this PR.

alexburke-abtion:
I was trying to remember my exact intent with this but haven’t been able to do so.

On that basis, I’m gonna say we close this for now - as pointed out by folks my description didn’t match the content very well, so at the very least reopening would mean having to rewrite it and it needs that anyway.

PR #1366, opened 2023-01-04
Use mailtrap by default

jeppester:
To make initial project setup easier.

Asana: https://app.asana.com/0/1142794766483633/1202242176448729

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 require "active_support/core_ext/integer/time"
+require "open-uri"
+require "json"
 
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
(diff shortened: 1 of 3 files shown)


config/environments/production.rb
```diff
@@ -59,16 +61,29 @@
...
+  if ENV.key?("SENDINBLUE_USERNAME")
+    config.action_mailer.smtp_settings = {
+      address: "smtp-relay.sendinblue.com",
+      port: 587,
+      user_name: ENV.fetch("SENDINBLUE_EMAIL"),
+      password: ENV.fetch("SENDINBLUE_PASSWORD"),
+      authentication: "login",
+      enable_starttls_auto: true
+    }
+  elsif ENV.key?("MAILTRAP_API_TOKEN")
```
jeppester, line 75: I would have preferred the app to break on boot when both of the keys are missing, but that unfortunately breaks our CI due to the asset precompilation step leading the production env settings.

config/environments/production.rb
```diff
@@ -59,16 +61,29 @@
...
+    response = URI.parse("https:///api/v1/inboxes.json?api_token=#{ENV.fetch('MAILTRAP_API_TOKEN')}").read
+    first_inbox = JSON.parse(response)[0]
```
hcarreras, line 77: Smart. Is this from the documentation?

hcarreras (approved):
I like the idea, and I approve of mailtrap.io 
It would also stay that in Staging Mailtrap could stay (although it would make a difference between staging and production, it's quite confusing to receive an email from a Staging environment)

config/environments/production.rb
jeppester: Yes, but I altered it slightly to not require a third party gem.

jeppester:
> I like the idea, and I approve of mailtrap.io It would also stay that in Staging Mailtrap could stay (although it would make a difference between staging and production, it's quite confusing to receive an email from a Staging environment)

I have changed the text slightly to include that sendinblue is usually set up only for the production env.

PR #1384, opened 2023-01-12
CI: Remove insecure db trust auth

runephilosof-abtion:
Postgres docs says:
>          You may also use "POSTGRES_HOST_AUTH_METHOD=trust" to allow all
 >         connections without a password. This is *not* recommended.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,7 +128,6 @@ jobs:
         env:
           POSTGRES_USER: postgres
           POSTGRES_PASSWORD: postgres
-          POSTGRES_HOST_AUTH_METHOD: trust
         ports:
           - 5432:5432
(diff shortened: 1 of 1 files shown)


jeppester (changes requested):
We need to update the readme example so that:
`postgresql://postgres@localhost:5432` instead becomes: `postgresql://postgres:postgres@localhost:5432`

jeppester:
I've never seen trust auth as an issue for local development DBs. If someone has access to my local machine, there are so many other ways to grab the data (like getting the pw from the docker-compose file). To be on the safe side I've just checked whether or not running the DB with docker compose exposes it externally and it does not.

For that reason I don't think we are in a hurry to implement this change on muffi using projects.

I don't think it's a bad change though. It has no impact on how easy the template is to use, but removes a line of configuration and possibly confusion about why that line was there.

jeppester:
I realize now that this is in CI. Lets get it merged right away :-)

PR #1385, opened 2023-01-12
CI: Update postgres to 14.6

runephilosof-abtion:
This is the default version used on Heroku currently

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,7 +124,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres:12
+        image: postgres:14.6
         env:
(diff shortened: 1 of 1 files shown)


jeppester (approved):
:+1:

jeppester:
We forgot to update the version in `app.json`. I'm going to do that directly on main.

I just got the `app.json` file in a working state again. It would be a shame for it to already become outdated!

jeppester:
And we also forgot the docker compose file. Updating that as well

jeppester:
Done

runephilosof-abtion:
I am adding 1SLA` label to this one, even though each project should of course match their production version

PR #1507, opened 2023-03-21
Implement specs' VCR labeling (through spec metadata)

davideluque:
A solution proposed by @hcarreras

We noticed in another project that VCR is global for all specs, but there are some specs where you don't want VCR to interfere.

Even though VCR is hooked into webmock, if you haven't mocked the request correctly VCR will take precedence and you won't be able to see the mocking error.

This solution proposes that specs should have vcr: true in their metadata, e.g.

`RSpec.describe MyController, type: controller, vcr: true`

Co-authored-by: hcarreras <hc@abtion.com>

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -72,4 +72,8 @@ RSpec.configure do |config|
   config.before do
     stub_strong_password
   end
+  config.around do |example|
+    vcr = example.metadata[:vcr]
+    vcr ? example.run : VCR.turned_off { example.run }
+  end
 end


davideluque:
Closed because our rails template does not have VCR as part of the base gems.

jeppester:
Do you think we should have it as a base gem?

We had it in the discuss board for a very long time, but we didn't discuss it much (due to the voting system) and it seemed like neither the people for or against were very passionate about the topic.

So in the end we decided to remove the card. That is - however - not to say that it can't be added again.

hcarreras:
We use it in VIR SAAF. I dislike the output format, but I have not found a better alternative.
Although I must say, I don't think it's worth adding a new gem at this stage.

PR #1611, opened 2023-05-15
Don't report sidekiq failures to rollbar until third retry

jeppester:
Asana:
https://app.asana.com/0/1142794766483633/1204338271508884/f

### DIFF
--- a/config/initializers/rollbar.rb
+++ b/config/initializers/rollbar.rb
@@ -50,6 +50,10 @@ Rollbar.configure do |config|
   # You can supply custom Sidekiq options:
   # config.use_sidekiq 'queue' => 'default'
 
+  # Don't report exceptions from retrying sidekiq jobs until third retry
+  # This is unrelated to and independent from `use_sidekiq` (though the docs are very unclear)
+  config.sidekiq_threshold = 3
+
   # If your application runs behind a proxy server, you can set proxy parameters here.
   # If https_proxy is set in your environment, that will be used. Settings here have precedence.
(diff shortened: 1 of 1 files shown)


config/initializers/rollbar.rb
```diff
@@ -50,6 +50,10 @@
   # You can supply custom Sidekiq options:
   # config.use_sidekiq 'queue' => 'default'
 
+  # Don't report exceptions from retrying sidekiq jobs until third retry
+  # This is unrelated to and independent from `use_sidekiq` (though the docs are very unclear)
+  config.sidekiq_threshold = 3
```
runephilosof-abtion, line 55: https://github.com/rollbar/rollbar-gem/blob/master/docs/configuration.md#sidekiq_threshold

> Ignored unless you've called `use_sidekiq`.

I think we should remove `sidekiq_threshold` again instead of enabling `use_sidekiq`.
I do not like how hard it is to figure out how [the sidekiq plugin](https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/plugins/sidekiq/plugin.rb) works.
Also, it will probably be unexpected with this threshold. Meaning that jobs that are not retried (for instance, because they are cron jobs running every 5 minutes) will never get retried.

runephilosof-abtion: I realize that I was the one championing for enabling this feature, and that jobs not retrying will get reported anyway...
So maybe we should enable `use_sidekiq` and keep the threshold. I can see both benefits and potential problems.

PR #1917, opened 2023-11-23
Add docker development setup

jeppester:
https://app.asana.com/0/1142794766483633/1204852357385065


config/database.yml
```diff
@@ -8,10 +8,10 @@ default: &default
...
 local: &local
   <<: *default
-  username: <%= ENV.key?("DATABASE_SERVER") ? URI(ENV.fetch("DATABASE_SERVER")).user : "" %>
-  password: <%= ENV.key?("DATABASE_SERVER") ? URI(ENV.fetch("DATABASE_SERVER")).password : "" %>
-  host: <%= ENV.key?("DATABASE_SERVER") ? URI(ENV.fetch("DATABASE_SERVER")).host : "" %>
-  port: <%= ENV.key?("DATABASE_SERVER") ? URI(ENV.fetch("DATABASE_SERVER")).port : "" %>
+  username: <%= ENV.key?("DATABASE_SERVER") ? URI(ENV.fetch("DATABASE_SERVER")).user : nil %>
+  password: <%= ENV.key?("DATABASE_SERVER") ? URI(ENV.fetch("DATABASE_SERVER")).password : nil %>
+  host: <%= ENV.key?("DATABASE_SERVER") ? URI(ENV.fetch("DATABASE_SERVER")).host : nil %>
+  port: <%= ENV.key?("DATABASE_SERVER") ? URI(ENV.fetch("DATABASE_SERVER")).port : nil %>
```
jeppester, line 14: If DATABASE_SERVER is set to "" as now mentioned in the README, each of these lines will be `nil`.

Maybe it's better with just
```erb
<%= URI(ENV.fetch("DATABASE_SERVER")).user %>
```

As the string is now always present :thinking:

package.json
```diff
@@ -26,7 +26,7 @@
     "css-loader": "^6.8.1",
     "css-minimizer-webpack-plugin": "^4.2.2",
     "dot-prop": "^7.2.0",
-    "mini-css-extract-plugin": "^2.7.2",
+    "mini-css-extract-plugin": "^2.7.6",
```
jeppester, line 29: I'm getting rid of this

yarn.lock
```diff
@@ -6672,10 +6672,10 @@ min-indent@^1.0.0:
...
   integrity sha512-I9jwMn07Sy/IwOj3zVkVik2JTvgpaykDZEigL6Rx6N9LbMywwUSMtxET+7lVoDLLd3O3IXwJwvuuns8UB/HeAg==
 
-mini-css-extract-plugin@^2.7.2:
-  version "2.7.2"
-  resolved "https://registry.yarnpkg.com/mini-css-extract-plugin/-/mini-css-extract-plugin-2.7.2.tgz#e049d3ea7d3e4e773aad585c6cb329ce0c7b72d7"
-  integrity sha512-EdlUizq13o0Pd+uCp+WO/JpkLvHRVGt97RqfeGhXqAcorYo1ypJSpkV+WDT0vY/kmh/p7wRdJNJtuyK540PXDw==
+mini-css-extract-plugin@^2.7.6:
+  version "2.7.6"
+  resolved "https://registry.yarnpkg.com/mini-css-extract-plugin/-/mini-css-extract-plugin-2.7.6.tgz#282a3d38863fddcd2e0c220aaed5b90bc156564d"
+  integrity sha512-Qk7HcgaPkGG6eD77mLvZS1nmxlao3j+9PkrT9Uc7HAE1id3F41+DdBRYRYkbyfNRGzm8/YWtzhw7nVPmwhqTQw==
```
jeppester, line 6678: And getting rid of this

.env
```diff
@@ -1 +1,7 @@
+REDIS_URL=redis://localhost:6379
+DATABASE_SERVER="postgresql://postgres@localhost:5432"
+SEED_ADMIN_EMAIL=admin@example.com
+SEED_ADMIN_INITIAL_PASSWORD=password
+SIDEKIQ_PASSWORD=password
+SIDEKIQ_USERNAME=abtion
```
jeppester, line 6: The `dotenv-rails` gem is only loaded for envs `development` and `test`.

So we can safely store shared vars in here without risking them showing up in production.

README.example.md
```diff
@@ -51,32 +53,83 @@ This section should include any business related explanation that helps understa
...
-```
+1. First install node and ruby:
+
+```bash
+# Add nodejs and ruby plugins if you don't already have them
+asdf plugin add nodejs
+asdf plugin ad ruby
+
+# Install
 asdf install
```
reinisla, line 79: Normally after I install a new nodejs version and the project is using yarn I have to run `npm -g install yarn` but it might be just my setup.

Also when I just run `asdf install` it will try to install some old stuff from other projects like 
python, java

```dotnet-core 6.0.400 is already installed
java zulu-jre-javafx-19.30.11 is already installed
Trying to update node-build... ok
Downloading node-v18.17.1-darwin-arm64.tar.gz...
-> https://nodejs.org/dist/v18.17.1/node-v18.17.1-darwin-arm64.tar.gz

WARNING: node-v18.17.1-darwin-arm64 is in LTS Maintenance mode and nearing its end of life.
It only receives *critical* security updates, *critical* bug fixes and documentation updates.

Installing node-v18.17.1-darwin-arm64...
Installed node-v18.17.1-darwin-arm64 to /Users/reinis/.asdf/installs/nodejs/18.17.1

python 3.10.5 is already installed
python 2.7.18 is already installed
ruby 3.2.2 is already installed
```

I assume you dont get such messages.
I normally install tool directly by running `asdf install nodejs 18.17.1`

README.example.md
```diff
@@ -51,32 +53,83 @@ This section should include any business related explanation that helps understa
...
+
+The steps are the same as for [Run locally using docker for database servers](#run-locally-using-docker-for-database-servers) except for step 2, and `docker` is (obviously) not a requirement.
+
+You will instead need to install the database servers and configure them correctly:
+
+- PostgreSQL (see version in docker-compose.yml)
+- Redis (see version in docker-compose.yml)
+
+It will be necessary to configure the database connections by setting corresponding env vars.
+If postgres is set up to use trust authentication (default for homebrew), then the following should be enough:
```
reinisla, line 106: Not sure people will know what is trust authentication maybe you should mention something if you are running `export OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES` command regulary or encountering `__NSCFConstantString initialize` errors in logs

README.example.md
```diff
@@ -95,16 +148,6 @@ If you need to make local changes to the env files, create a `.env.ENVIRONMENT.l
```
reinisla, line 148: maybe add a comment about just plain .env.local

spec/support/capybara.rb
```diff
@@ -20,6 +20,8 @@
 RSpec.configure do |config|
   config.before(:each, type: :system) do
     driven_by(:selenium, using: ENV.fetch("CAPYBARA_DRIVER", "headless_chrome").to_sym,
-                         screen_size: [1920, 1080])
+                         screen_size: [1920, 1080]) do |capabilities|
+      capabilities.add_argument("--no-sandbox") if ENV["CAPYBARA_NO_SANDBOX"].present?
```
reinisla, line 24: what does this do?

README.example.md
jeppester: Believe it or not; I actually changed this line entirely, but then realised it was out of scope for this PR and reverted it :grinning:

jeppester: I think I'll just make that change directly in the main branch.

spec/support/capybara.rb
jeppester: I had to add it because otherwise chromium will fail to run inside the dev container.
Thus it's only enabled in the `Dockerfile` and can otherwise be ignored.

README.example.md
jeppester: I was thinking about making this line more `homebrew`-oriented.
So something like:
```md
If you've installed postgres and redis with homebrew (e.g. postgres is using trust authentication), then the following should be enough:
```

README.example.md
jeppester: > Normally after I install a new nodejs version and the project is using yarn I have to run npm -g install yarn but it might be just my setup.

Great point, I'll add that :+1: 

> Also when I just run asdf install it will try to install some old stuff from other projects like
python, java

I think it lists other tool chains because you have them in a .tool-versions somewhere up the folder hierarchy, could be `~/.tool-versions`.

README.example.md
jeppester: I've updated this in main

README.example.md
reinisla: For me it sounds the same as what is written, the issue is, that it does not tell me why should I set the DATABASE_SERVER to ""

What if I skipped this step because I thought that it was irrelevant for me, then I see errors and I would just run `export OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES` as I am used to. 
This is just my perspective on how things would play out

reinisla (approved):

README.example.md
jeppester: Is it better now?

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+RUN chmod +x wait-for-sigterm.sh
+
+# Switch to correct work dir
+WORKDIR /app
+
+# ASDF
+COPY .tool-versions ./
+RUN asdf plugin add nodejs
+RUN asdf plugin add ruby
+ENV PATH=$PATH:/root/.asdf/shims:/root/.asdf/bin
```
jeppester, line 37: This line must be moved above the asdf calls

README.example.md
davideluque: Chiming in here:

`OBJC_DISABLE_INITIALIZE_FORK_SAFETY` is not related in any way to Docker or Postgres. It is an issue with Ruby and MacOS.

docker-compose.yml
```diff
@@ -13,5 +13,47 @@ services:
...
+    image: dev
+    build: .
+    command: bash -c "asdf install; npm install -g yarn; bundle install; yarn"
+    working_dir: /app
+    volumes:
+      - .:/app
+      - node-modules:/app/node_modules
+      - asdf:/root/.asdf
+
+  dev:
```
hcarreras, line 29: why does it know that this container should not start when you run `docker compose up`?

hcarreras (commented):
I like the flexibility while keeping the complexity low.

docker-compose.yml
jeppester: Because it belongs to the "dev" profile.

You will have to run `docker compose --profile dev up`. E.g. add the specific profile to the call.

README.example.md
jeppester: @davideluque Thank you for participating.

I'm aware that the issue is not really the template's or postgres' fault. Still it seems that our devs will likely face that specific issue, which is difficult to debug.

I am trying to be pragmatic here, without spending too much precious README space on it.

Let me know if you have a suggestion for how we can better solve this problem.

docker-compose.yml
hcarreras: Cool. I did not know this compose feature

jeppester:
@runephilosof-abtion Please have a look :-)

README.example.md
```diff
@@ -51,32 +53,88 @@ This section should include any business related explanation that helps understa
...
+
+It will be necessary to configure the database connections by setting corresponding env vars.
+If postgres is set up to use trust authentication (default for homebrew), then the following should be enough\*:
+
+```bash
+# .env.local
+DATABASE_SERVER=""
 ```
 
+\*_Caution: If you on a mac do not clear `DATABASE_SERVER`, you might get a seemingly unrelated: `+[__NSCFConstantString initialize] may have been in progress in another thread...`_
```
runephilosof-abtion, line 116: Seems odd why it would do a difference, since it defaults to `""`, when the variable is not set.
> host: <%= URI(ENV.fetch("DATABASE_SERVER", '')).host %>

runephilosof-abtion (approved):
Looks good. I have a comment on `asdf` versus `ruby:3.x` image.
But it is not blocking, just a remark.

README.example.md
jeppester: This PR adds a default `DATABASE_SERVER` to the template (in `.env`) so that docker will work out of the box.
Several apple silicon mac-devs have faced the mentioned ruby error on other projects where we've done the same. It doesn't matter if the connection string is also correct for the local postgres server.

I added a caution-note to hopefully prevent us from wasting further time on this bug.

When the connection settings are `nil` (and according to @reinisla's testing also when they are `""`) rails will - to my best knowledge - connect to postgres through a socket, whereas it will otherwise use TCP/IP. I think it is this difference that works around the ruby bug.

README.example.md
davideluque: Hey @jeppester 

I think it could be misleading to connect the NSFC error with the connection string / docker. This error is related to MacOS and Multithreading, as someone posted on StackOverflow:

> Screenshot 2023-11-30 at 09 15 23

davideluque: Here's the link where someone experienced it in Python: https://stackoverflow.com/questions/50168647/multiprocessing-causes-python-to-crash-and-gives-an-error-may-have-been-in-progr

reinisla: The fact is that by changing the DATABASE_SERVER="" the error goes away

jeppester: To my best knowledge the situation is:
1. When using a socket to connect to a local postgres there is no issue
2. When using tcp/ip to connect to a local postgres the error happens

Emptying `DATABASE_SERVER` it will lead to 1, doing nothing will lead to 2.

What I'm trying to do here is to help our devs work around that problem rather having to figure out this illogical work-around by themselves.

What would you suggest that we do?

jeppester:
Perfect is the enemy of good. I'm merging this. We can improve upon it later.

README.example.md
runephilosof-abtion: So you need to overwrite the `DATABASE_SERVER` set in `.env`.
In that case, I think we should move `DATABASE_SERVER` from `.env` to `docker-compose.yml`.

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN chmod +x wait-for-sigterm.sh
+
+# Switch to correct work dir
+WORKDIR /app
+
+# ASDF
+ENV PATH=$PATH:/root/.asdf/shims:/root/.asdf/bin
+COPY .tool-versions ./
+RUN asdf plugin add nodejs
+RUN asdf plugin add ruby
```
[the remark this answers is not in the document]
runephilosof-abtion: 1. When I upgrade Ruby, I change it in Dockerfile and run `docker compose build dev`. So no, it is not needed when upgrading.

README.example.md
jeppester: >  In that case, I think we should move DATABASE_SERVER from .env to docker-compose.yml.

That is a great idea!

jeppester: I realize now that setting the env var in the Docker compose file will not work for the hybrid setup that most of our devs are using. E.g. running DBs in docker, but the rails app locally. Unless I'm missing something here?

runephilosof-abtion: Right, so we are already setting it in both `.env` (for "hydrid") and overriding it in `docker-compose.yml` (for all docker).
This section says that people using no docker at all, should override the `.env` one with an empty `DATABASE_SERVER`. :+1: 

I have created https://github.com/abtion/rails-template/issues/1941 to discuss this line further.

PR #1980, opened 2024-02-01
RackAttack: Enhanced security with updated rate limits

davideluque:
Expanded the Rack::Attack configuration to tighten security by refining throttling rules and rate limits to protect against brute-force attacks and DDoS attempts. Adjusted the general IP-based request limit to 300 requests every 5 minutes, significantly enhancing defenses against high-volume traffic abuses. Also introduced specific throttles on login attempts by IP and email address to mitigate brute-force login attacks, capped at 5 requests within a 20-second window. Updated tests to align with the new rate limit configurations, ensuring robust validation of the protective measures. The changes prepare the application for scenarios involving high traffic and potential malicious activities without affecting legitimate users' access under normal conditions.

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -30,7 +30,10 @@ module Rack
     # Throttle all requests by IP (60rpm)
     #
     # Key: "rack::attack:#{Time.now.to_i/:period}:req/ip:#{req.ip}"
-    throttle("req/ip", limit: 300, period: 5.minutes, &:ip)
+    throttle("req/ip",
+             limit: ENV.fetch("RACK_ATTACK_THROTTLE_LIMIT", "300").to_i,
+             period: ENV.fetch("RACK_ATTACK_THROTTLE_PERIOD", "5").to_i.minutes,
+             &:ip)
 
     ### Prevent Brute-Force Login Attacks ###
 
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -1,5 +1,82 @@
 # frozen_string_literal: true
 
-Rack::Attack.throttle("requests by ip", limit: 20, period: 2, &:ip)
-Rack::Attack.blocklist_ip(ENV.fetch("BLOCKED_IPS")) if ENV.fetch("BLOCKED_IPS", nil).present?
-Rack::Attack.enabled = Rails.env.production? || ENV.key?("ENABLE_RACK_ATTACK")
+module Rack
+  class Attack
+    Rack::Attack.enabled = Rails.env.production? || ENV["ENABLE_RACK_ATTACK"].present?
+
+    Rack::Attack.blocklist_ip(ENV["BLOCKED_IPS"]) if ENV["BLOCKED_IPS"].present?
+
+    ### Configure Cache ###
+
+    # If you don't want to use Rails.cache (Rack::Attack's default), then
+    # configure it here.
+    #
+    # Note: The store is only used for throttling (not blocklisting and
+    # safelisting). It must implement .increment and .write like
+    # ActiveSupport::Cache::Store
+
+    # Rack::Attack.cache.store = ActiveSupport::Cache::MemoryStore.new
+
+    ### Throttle Spammy Clients ###
+
+    # If any single client IP is making tons of requests, then they're
+    # probably malicious or a poorly-configured scraper. Either way, they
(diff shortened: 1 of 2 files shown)


spec/rack/attack_spec.rb
```diff
@@ -5,56 +5,103 @@
...
-    # Use memory cache store for this test
     memory_store = ActiveSupport::Cache.lookup_store(:memory_store)
     allow(Rails).to receive(:cache).and_return(memory_store)
     allow(Rack::Attack).to receive(:enabled).and_return(true)
     Rack::Attack.reset!
   end
 
   describe "IP address throttling" do
-    let(:limit) { 20 }
+    let(:limit) { 300 }
```
jeppester, line 27: I'm not against making this change in the configuration, but I fear that it will make the test unbearably slow. I wonder if there is a shortcut we can take in order to still test, but faster.

Ideas on top of my head:
- Force to cache key to be close to the limit
- Use another type for the spec that does not exercise as much of the middle ware (I would have to investigate more to know if such a thing exists)

A third solution could be to drop the spec completely. I'm just not very happy about implementing something in the code that we have no clue if works as intended (which is why I added this test in the first place)

davideluque:
Upgrade to Ruby 3.1+ is required for the testing suite to pass.

spec/rack/attack_spec.rb
davideluque: Valid point. I will explore solutions that keep the test as close to real-life scenarios as possible. Due to the passing tests, I thought Rack::Attack was enabled but we did not have a cache implemented so our tests did not reflect the real-life scenario.

runephilosof-abtion: this branch
```

```

main branch
```
```

runephilosof-abtion: this branch
```
  Rack::Attack
    8.42 seconds average (42.1 seconds / 5 examples) ./spec/rack/attack_spec.rb:5
```

main branch
```
  Rack::Attack
    0.82308 seconds average (2.47 seconds / 3 examples) ./spec/rack/attack_spec.rb:5
```

davideluque: 10 times slower is not doable, for sure .

spec/rack/attack_spec.rb
```diff
@@ -6,55 +6,76 @@
...
-            expect(response).to_not(have_http_status(:too_many_requests))
-          end
-        end
+    let(:limit) { 300 }
+
+    it "throttles requests from the same IP with status 429 when limit is exceeded" do
+      freeze_time do
+        # Get close to the limit without excersizing the whole stack for each request.
+        # Otherwise the spec would be really too slow
+        (limit - 1).times { Rack::Attack.cache.count("req/ip:1.2.3.5", 5.minutes) }
```
jeppester, line 22: This was the least "invasive" way I could find for mocking the request count.

jeppester:
@davideluque @runephilosof-abtion:
I think I've managed to make the spec fast without decreasing the test coverage.
Could you have another look?

It would be great to finally get this merged :-)

davideluque:
@jeppester Looks great to me :)

spec/rack/attack_spec.rb
runephilosof-abtion: The new solution for testing limit 300 is faster than main branch now
main branch `2.18 seconds`
now `0.54818 seconds`

config/initializers/rack_attack.rb
```diff
@@ -1,5 +1,82 @@
...
+    # don't deserve to hog all of the app server's CPU. Cut them off!
+    #
+    # Note: If you're serving assets through rack, those requests may be
+    # counted by rack-attack and this throttle may be activated too
+    # quickly. If so, enable the condition to exclude them from tracking.
+
+    # Throttle all requests by IP (60rpm)
+    #
+    # Key: "rack::attack:#{Time.now.to_i/:period}:req/ip:#{req.ip}"
+    throttle("req/ip", limit: 300, period: 5.minutes, &:ip)
```
runephilosof-abtion, line 33: It would be nice to be able to vary this based on whether you are authenticated.

Some apps might have javascript polling each second or similar.
For companies using vpn, a lot of users would come from the same ip.

runephilosof-abtion: Should we limit this throttling to post?

PR #2044, opened 2024-04-30
Replace yarn with npm

jeppester:
https://app.asana.com/0/1142794766483633/1206382290441397


jeppester:
In order to check if npm is still fast with many deps I replaced yarn with npm on one of our nextjs projects:
https://github.com/abtion/dsvt-portal/pull/11

npm is also faster than yarn on that project.

For that reason I believe I can safely say that switching to npm won't be a big deal.

PR #2144, opened 2024-06-11
Rubocop: Remove TargetRubyVersion

runephilosof-abtion:
Less places to update when upgrading.
It will use the version from `.ruby-version`

https://docs.rubocop.org/rubocop/configuration.html#setting-the-target-ruby-version
> If a TargetRubyVersion is not specified in your config, then RuboCop will check your project for a series of other files where the Ruby version may be specified already. The files that will be checked are (in this order): *.gemspec, .ruby-version, .tool-versions, and Gemfile.lock.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,7 +7,6 @@ AllCops:
   DisplayCopNames: true
   DisplayStyleGuide: true
   NewCops: enable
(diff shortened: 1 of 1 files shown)


jeppester (approved):
Nice!

PR #2145, opened 2024-06-11
Only enable `enable_reloading` if using Spring

runephilosof-abtion:
Only enable `enable_reloading` if run through `bin/spring rspec`

`enable_reloading` is the new naming for that configuration option.
By default it is `false` in test.rb for a new Rails app.
Spring requires it `true`, so set it true when Spring is being used.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -10,7 +10,7 @@ require "active_support/core_ext/integer/time"
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
-  config.cache_classes = false
+  config.cache_classes = !defined?(Spring)
 
   # Do not eager load code on boot. This avoids loading your whole application
   # just for the purpose of running a single test. If you are using a tool that
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -10,7 +10,7 @@ require "active_support/core_ext/integer/time"
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
-  config.cache_classes = !defined?(Spring)
+  config.enable_reloading = defined?(Spring)
 
   # Do not eager load code on boot. This avoids loading your whole application
   # just for the purpose of running a single test. If you are using a tool that


jeppester:
What would be the issue with just enabling cache_classes in general?

It seems to me that it should only be a problem when running specs in watch mode (with something like guard)

runephilosof-abtion:
I can see my description is off, partly because I assumed `cache_classes = true` before my change.
But I can see I changed it to false for Spring in https://github.com/abtion/rails-template/commit/be366ba8c72adc650bc917ba26092e2a98a883ab  Oct 13th, 2021...
I have updated the description.

I assume `cache_classes = true` is faster, since that is the default in the Rails generator.
(Actually, the setting has been renamed to `enable_reloading = false` in Rails 7.1.)

I will do some more tests on what the benefits of setting it to `true` is.

runephilosof-abtion:
I ran a couple of CI runs in a bigger project

| `cache_classes = true` | `cache_classes = false` |
| ---------------- | ---------------- |
| 3.13 | 3.28 |
| 3.11 | 3.19 |
| 4.10 | 4.22 |
| 3.03 | 3.34 |

Hard to say with that variance. Might be around 5%. Would need to run it in an environment with less variance.

runephilosof-abtion:
> Hard to say with that variance. Might be around 5%. Would need to run it in an environment with less variance.

I should have also kept the seed the same.

jeppester:
> > Hard to say with that variance. Might be around 5%. Would need to run it in an environment with less variance.
> 
> I should have also kept the seed the same.

That is a good point, but it does seems like a larger boost than what could be explained by randomness.

jeppester:
I think I'll do a few tests on my server as well, I think it might have less variance due to not having UI apps running

jeppester:
I have now tested on the template itself and two other projects. Interestingly `cache_classes` (or `enable_reloading`) was already true for the two other projects.

- Template: 10 tests, test suite takes ~3 seconds, `cache_classes=false` is 5% slower
- Project A: 20 tests, test suite takes ~6 seconds, `cache_classes=true` is 1,5% slower
- Project B: 10 tests, test suite takes ~25 seconds, no difference

I don't think the difference is significant on my setup.
Still I think we should rename the setting to `enable_reloading` and I also think it makes the most sense to set it to `false` for our tests (unless if it has a negative impact on spring?)

runephilosof-abtion:
I actually had a lot of failing specs (randomly which of the specs in the suite each time) on a project because of this.
The problem is that `dotenv-rails` [loads Spring](https://github.com/bkeepers/dotenv/blob/b86d9e4c23a9f1bc83f8853f6a1b244449c65cd0/lib/dotenv/rails.rb#L14), if it is able to (rescues if not).
The feature specs became slow enough to fail (worked if increasing `Capybara.default_max_wait_time = 5`).
So in that project, I have removed Spring from the Gemfile, until I figure out a solution.

PR #2191, opened 2024-06-26
Update app.json: postgresql plan mini=>essential-0

runephilosof-abtion:
Mini is no more.


jeppester (approved):
> Mini is no more.

Rest in peace :candle:

PR #2201, opened 2024-07-01
Gemfile: Gems out of test env & `require: false`

runephilosof-abtion:
They do not have rake tasks or similar code used in development env. Less code to load, means faster booting time for development env (very little).

Brakeman is always executed using bundle exec brakeman or using its binstub, which will still work with `require: false` presidentbeef/brakeman#1854


jeppester (approved):
Great :clap:

PR #2202, opened 2024-07-01
CI auto-merge-dependabot: Only run on dependabot branches

runephilosof-abtion:


jeppester (approved):
Great idea :+1:

jeppester:
@runephilosof-abtion
FYI: This change caused automerge to only run on PRs that were targeting branch starting with "dependabot/".
I've now made the following change which appears to make it work as intended:
https://github.com/abtion/rails-template/commit/6f3164e06f08602fd42752d2cba4ed7fff1cea85

Based on the last paragraph in this section:
https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-pull_request_target-workflow-based-on-the-head-or-base-branch-of-a-pull-request

PR #2328, opened 2024-09-11
CI: Annotate errors for Github

runephilosof-abtion:
As described on
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions

PR #2329, opened 2024-09-12
Make it fail

runephilosof-abtion:
Test PR for #2328


jeppester:
Nice !

runephilosof-abtion:
Test completed, see the files changed for the annotations.
It would have been nicer if the place where errors were raised had been annotated, but better than no annotations.

PR #2330, opened 2024-09-12
CI: comment with rspec-stderr

runephilosof-abtion:
We should avoid introducing noisy warnings in our test output.
Often the warnings are added by Dependabot upgrades. This way it will not merge PR adding noise automatically.

https://app.asana.com/0/1142794766483633/1208315235207762


runephilosof-abtion:
This will probably include the annotations from #2328 in the comment...

runephilosof-abtion:
> This will probably include the annotations from #2328 in the comment...

It looks like it is using `stdout`, so it won't be a problem.
https://github.com/Drieam/rspec-github/blob/075c231617113d9a05c80513dff17c7c7e27dbae/lib/rspec/github/formatter.rb

runephilosof-abtion:
Optimally, not only from the rspec step.

Also, it would be nice to fix that warning before merging this.

runephilosof-abtion:
Maybe I should make the step fail after posting, so that Dependabot PRs are not auto merged.

runephilosof-abtion:
Oops. All the steps are being run twice (push and pull_request), wasting resources.

runephilosof-abtion:
I will test it on another project first. Turning it into a draft.

jeppester (commented):
This is a really interesting idea.
I do wonder if people will learn to ignore this message that same way as normal terminal output.

I wish there was a way to print only the additional error output added by the PR's branch

runephilosof-abtion:
@jeppester 
> I do wonder if people will learn to ignore this message that same way as normal terminal output.

Now, it cannot be ignored. Since it will fail the CI.

jeppester:
@runephilosof-abtion How would you feel about having the exceptions (e.g. "Stopped processing SimpleCov as a previous error not related to SimpleCov has been detected")  in a separate file that we would be able to use as an escape hatch?

We could place the file alongside the CI-files to not clutter the rest of the source code.

I have time to implement it today if you are not against that idea.

PR #2333, opened 2024-09-16
README.example.md: Update observatory badge link

runephilosof-abtion:
https://developer.mozilla.org/en-US/blog/mdn-http-observatory-launch/
> The [old Mozilla Observatory](https://observatory.mozilla.org/) ... has been deprecated and will be sunset in September 2024.

I wonder what will happen when they "sunset" it. Will the old link redirect to the new http observatory?

Let's wait with merging this until we see the "sunset".

### DIFF
--- a/README.example.md
+++ b/README.example.md
@@ -1,6 +1,6 @@
 # Project Name Human
 
-[![Observatory](https://img.shields.io/mozilla-observatory/grade-score/abtion-rails-template.herokuapp.com)](https://observatory.mozilla.org/analyze/abtion-rails-template.herokuapp.com)
+[![Observatory](https://img.shields.io/mozilla-observatory/grade-score/abtion-rails-template.herokuapp.com)](https://developer.mozilla.org/en-US/observatory/analyze?host=abtion-rails-template.herokuapp.com)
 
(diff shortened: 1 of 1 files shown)


jeppester:
> I wonder what will happen when they "sunset" it. Will the old link redirect to the new http observatory?

The main reason for the move is that Mozilla wants observatory to be part of MDN, so I think it's safe to assume that the new endpoint is what they would want us to use going forward.

For that reason I don't see a reason not to merge this.

runephilosof-abtion:
Another reason to wait.

[It](https://developer.mozilla.org/en-US/blog/mdn-http-observatory-launch/#has_the_http_observatory_api_been_updated_to_use_the_new_tests) also says 
> Not yet. The API will continue using the old test infrastructure for a while, therefore you will see some small differences between test scores returned by the API and the website. The API will be updated to use the new tests in a near-future iteration.

So there might be some inconsistency between the scores.

jeppester:
> So there might be some inconsistency between the scores.

This is a problem that will solve itself in (hopefully short) time, and the implications are near non-existent.
I'd rather get this merged and not have it stuck in my mental todo-list :smile:

PR #2381, opened 2024-10-02
Upgrade webpack

runephilosof-abtion:
https://github.com/abtion/rails-template/security/dependabot/77
> Cross-realm object access in Webpack 5


runephilosof-abtion:
@jeppester I don't know why, but dependabot was not able to make this update. It claimed that our settings would not allow it to update `webpack`, which does not make sense to me.

PR #2607, opened 2025-01-20
Use redis for caching

jeppester:
https://app.asana.com/0/1142794766483633/1208101194508777


config/environments/production.rb
```diff
@@ -68,7 +68,10 @@
   config.lograge.enabled = true
 
   # Use a different cache store in production.
-  # config.cache_store = :mem_cache_store
+  config.cache_store = :redis_cache_store, {
+    url: ENV.fetch("REDIS_URL", "redis://localhost:6379"),
+    expires_in: 1.week
+  }
```
runephilosof-abtion, line 74: On Heroku you need to either import their certificate somehow or as they suggest not verify
```
      ssl_params: { verify_mode: OpenSSL::SSL::VERIFY_NONE }
```
```suggestion
  config.cache_store = :redis_cache_store, {
    url: ENV.fetch("REDIS_URL", "redis://localhost:6379"),
    expires_in: 1.week,
    ssl_params: { verify_mode: OpenSSL::SSL::VERIFY_NONE }
  }
```

runephilosof-abtion:
You need to use redis premium to get somewhat reliable response times on Heroku (which is quite expensive)

jeppester:
> You need to use redis premium to get somewhat reliable response times on Heroku (which is quite expensive)

Good point. I'll close this.

